PR #67, opened 2015-01-07
template makeQuote disable-output-escaping

bwbohl:
$preQuote and $postQuote values should be inserted with
disable-output-escaping="yes" in order to allow submission of unicode
values


bwbohl:
Of course I can out ordinary unicode characters in the variables. I'd regard it more as a feature, being able to insert unicode codepoints. This will not break if I insert a "ordinary" character but it will also output the intended result if I input a codepoint.

sebastianrahtz:
I am still not getting this. if you want to put &#x2345; in the variable,
whats the problem?

On 8 January 2015 at 09:50, Benjamin W. Bohl <notifications@github.com>
wrote:

> Of course I can out ordinary unicode characters in the variables. I'd
> regard it more as a feature, being able to insert unicode codepoints. This
> will not break if I insert a "ordinary" character but it will also output
> the intended result if I input a codepoint.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/TEIC/Stylesheets/pull/67#issuecomment-69157526>.
>



-- 

Sebastian Rahtz

Director (Research) of Academic IT

University of Oxford IT Services

13 Banbury Road, Oxford OX2 6NN. Phone +44 1865 283431


Não sou nada.

Nunca serei nada.

Não posso querer ser nada.

À parte isso, tenho em mim todos os sonhos do mundo.

bwbohl:
at the moment ```&#x00BB;``` will be outputted as such in the PDF, ergo a string '```&#x00BB;```

sebastianrahtz:
surely &#x00BB; in your XSL is parsed by the XML processor as soon as the file is loaded, and managed internally in Unicode? I really don't see how it can get to the PDF output as the string &#x00BB. Entities must be resolved by the XML parser.

martindholmes:
I can confirm that if you insert numerical entity references as values of these two variables in Oxygen's TEI-to-PDF transformation scenario, they are interpreted as literal strings and the entity reference, rather than the character it's supposed to represent, is shown in the output.

scstanley7:
We agree that codepoints being output literally instead of being read as character references is annoying, but we think that it's easier to simply insert the literal character than mess around with `@disable-output-escaping`, so we're going to close without fixing this. Thank you for bringing this up, though!

Issue #73, opened 2015-01-12
FO line-height configuration


bwbohl:
Currently transforming TEI to FO and PDF doesn't allow specifying line-height. I consider implementing corresponding independent parameters for front, body and back.
If I set the line-height attribute in the corresponding page-sequences this will affect all content in the corresponding fo:flow, including footnotes. Besides footnotes cit[@rend='block'] might be another case where increased line-height is rather unwanted.
So alternatively:
- line-heights could either be set on div or p elements (excluding such things as footnotes and citations), or
- other parameters might be introduced to configure line-height for footnotes and cit[@rend='block'] respectively

Any thoughts or opinions?

bwbohl:
lists of course might be considered differently, too!

sebastianrahtz:
that's hard. my inclination is to do by inclusion, not exclusion. so apply
it to contexts where you know it makes sense,
eg on <p>, and add to other elements as you see the need.

On 12 January 2015 at 10:21, Benjamin W. Bohl <notifications@github.com>
wrote:

> Currently transforming TEI to FO and PDF doesn't allow specifying
> line-height. I consider implementing corresponding independent parameters
> for front, body and back.
> If I set the line-height attribute in the corresponding page-sequences
> this will affect all content in the corresponding fo:flow, including
> footnotes. Besides footnotes cit[@rend <https://github.com/rend>='block']
> might be another case where increased line-height is rather unwanted.
> So alternatively:
>
>    - line-heights could either be set on div or p elements (excluding
>    such things as footnotes and citations), or
>    - other parameters might be introduced to configure line-height for
>    footnotes and cit[@rend <https://github.com/rend>='block'] respectively
>
> Any thoughts or opinions?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/TEIC/Stylesheets/issues/73>.
>



-- 

Sebastian Rahtz

Director (Research) of Academic IT

University of Oxford IT Services

13 Banbury Road, Oxford OX2 6NN. Phone +44 1865 283431


Não sou nada.

Nunca serei nada.

Não posso querer ser nada.

À parte isso, tenho em mim todos os sonhos do mundo.

PR #81, opened 2015-02-12
add parameters for front, body, back line-height

bwbohl:
as discussion in #73 tended towards an inclusive rather than an exclusive approach,at the moment, this feature will only be applied to tei:p[not(parent::tei:note)]


bwbohl:
I just discovered some problems:
1) as <fo:footnote> might be a child of a <tei:p>'s <fo:block> the line height will also be applied to the footnote...

2) same might apply for block-quotations

I'll try to hand in some fixes for this.

PR #85, opened 2015-03-12
Trim leading whitespace in instrText, which may or may not be present.

jure:
So my Word document has this element:

```
      <w:r w:rsidR="00E86F6B">
        <w:instrText>HYPERLINK \l "_ENREF_18" \o "Temporary, 2003 #8947"</w:instrText>
      </w:r>
```

Which means that the current method of checking only for " HYPERLINK" fails, and produces several hyperlinks in the end, as each space separated entry gets converted into a separate link. The example above will result in something like (after Docx to TEI to XHTML):

```
[<a class="link_ref" href="HYPERLINK">18</a>, <a class="link_ref" href="\l">18</a>, <a class="link_ref" href="&#34;_ENREF_18&#34;">18</a>, <a class="link_ref" href="\o">18</a>, <a class="link_ref" href="&#34;Temporary,">18</a>, <a class="link_ref" href="2003">18</a>, and 18]
```

I've tried to find a reference in the standard that a leading space in `instrText` is mandatory, but it seems like it can either be there or not. Happy to be educated on that. Point remains, that my document has an `instrText` with a `HYPERLINK` without a leading space.

So when we fall back to `processInstruction`, perhaps there's no real harm in stripping the leading space, if it exists? The tokenization changes too, but other things remain the same.

 Let me know what you think.

### DIFF
--- a/docx/from/functions.xsl
+++ b/docx/from/functions.xsl
@@ -223,19 +223,22 @@ of this software, even if advised of the possibility of such damage.
 
   <xsl:function name="tei:processInstruction"  as="xs:string">
     <xsl:param name="instr"/>
+    <xsl:variable name="instr">
+      <xsl:value-of select="replace($instr, '^\s+|\s+$', '')"></xsl:value-of>
+    </xsl:variable>
     <xsl:choose>
       <xsl:when test="matches($instr,'REF _')"> <!-- this will also catch NOTEREF _ -->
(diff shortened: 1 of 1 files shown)


sebastianrahtz:
your code looks much safer, I agree, and I can't see a downside.

jure:
Thanks for the merge!

PR #88, opened 2015-03-20
added jTEI profile with jTEI-to-ODT transformation

rvdb:


lb42:
There seems to be a problem with tables in which the @cols attribute is used to get spanning cells. If *all* the cells in a row are spanned (as in the example in the documentation) this works OK, but if only some of them are, the other columns are suppressed.

PR #90, opened 2015-03-24
-don't copy /TEI/@rend[. = ('jTEI', 'jTEI.internal')] attribute

rvdb:


sebastianrahtz:
Can you be more specific? copy it to where? what behaviour do you expect?

rvdb:
We're using the /TEI/@rend attribute with possible values 'jTEI' and 'jTEI.internal' to associate jTEI articles with the jTEI document type we've defined in the jTEI framework in Oxygen. When transforming jTEI source articles to the [OpenEdition TEI format] (http://lodel.org/ns/tei.openedition.1.3/) used for the journal's [publication system] (http://revues.org), we don't want to copy that attribute, since we don't want to associate OpenEdition documents with the jTEI document type.

PR #109, opened 2015-06-15
Fix missing first text node, a [, in certain references.

jure:
I can provide an example doc, where I compiled the few forms of references I've noticed in my set, this fixes the issue for me, where a `hi` element actually containing text was ignored in certain situations because of this:
https://github.com/TEIC/Stylesheets/blob/master/docx/from/pass2.xsl#L330

Can you think of a better way?

Here's the example Docx, if you're interested: https://dl.dropboxusercontent.com/u/30920/referenceexamples.docx

Look for square brackets getting lost in the current conversion, it's most visually obvious in the Docx to HTML conversion, but is already present in the TEI.

### DIFF
--- a/docx/from/pass2.xsl
+++ b/docx/from/pass2.xsl
@@ -599,13 +599,15 @@ of this software, even if advised of the possibility of such damage.
 	  <xsl:value-of select="$target"/>
 	</xsl:processing-instruction>
 	<xsl:choose>
-          <xsl:when test="(count(.//text()) &lt; 2) or (count(tei:ref) &gt; 1)">
+          <xsl:when test="(count(.//text()) &lt; 3) or (count(tei:ref) &gt; 1)">
             <xsl:apply-templates mode="pass2"/>
           </xsl:when>
           <xsl:otherwise>
-            <xsl:value-of select="(.//text())[1]"/>
+            <xsl:value-of select="./text()[1]|(.//node()[not(self::tei:ref) and self::text()])[1]"/>
             <xsl:apply-templates select="*" mode="pass2"/>
-            <xsl:value-of select="remove(text(), 1)"/>
+            <xsl:if test="./text()[last()]">
+              <xsl:value-of select="(.//node()[not(self::tei:ref) and self::text()])[last()]"/>
+            </xsl:if>
           </xsl:otherwise>
         </xsl:choose>
       </xsl:when>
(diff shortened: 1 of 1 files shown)


jure:
I've found another exception which converts a `[4,5]` reference to `[4,55` with the new code, so I still need to fix this one. Oh when will the exceptions stop popping up? :)

jure:
You're going to think I'm crazy, but I can say that I've solved my known reference problems so far (and there are a lot!) thusly:
https://github.com/Tahi-project/Stylesheets/commit/f4e2f3f0da10ef8d6f6bd0a4ce5a27afbf12e975

I'm updating this PR with that additional code, for you to try it out.

jure:
Here you go: [referenceexamples.docx](https://github.com/TEIC/Stylesheets/files/3157796/referenceexamples.docx)

raffazizzi:
We have found the cause of the issue, but we're still working on a fix. 

A series of `tei:hi` elements with the same `@style and @rend` are merged together by the Stylesheet by picking the first one with some content and looping through the following ones. In the case submitted in this PR the first `tei:hi` is empty, therefore the following `tei:hi` with `[` gets dropped. We tried adjusting the condition that determines when a following `tei:hi` gets dropped to only process nodes with content, but we made it too restrictive so one of our tests (test29.docx) now fails as one `tei:hi` gets duplicated.

ebeshero:
@raffazizzi A thought, following our session this afternoon. There's a lot of labored XSLT processing dedicated to checking the contents of a previous-sibling `<hi>` generated during a first stage of processing (such that, if the previous sibling has content, don't process the current node, etc). If the goal in this Stylesheet is simply to suppress empty `<hi>` elements, why can't we simplify the processing? Can we simply check if the current context is empty and if so, don't process? (And if we can't why not?) 

And a note for future testing. The difference between the OP's Word document and ours in test29.docx is that our test29 involves a w:instrText that seems to be leading to a noteRef (it's a note that links elsewhere in the document). When we implemented our suggested fix yesterday, it corrected the OP's problem, but generated *duplicate* output, a second instance of the footnote number, 3. We need to figure out what causes this duplicate processing, as well as to find a way to simplify the suppression of empty `<hi>` elements.

raffazizzi:
@ebeshero @tuurma and @raffazizzi  fixed this in https://github.com/TEIC/Stylesheets/commit/2438e944bd03bd53e859eae5f4659a7c0675bbf8 (at Graz 2019 F2F meeting)

PR #114, opened 2015-08-12
regularise display of attribute values

stuartyeates:


hcayless:
Stuart, can you tell me what this is fixing and what the change will mean? It's kind of hard to tell just from looking at the diff. I can see you're simplifying stuff, but not what the implications are.  Thanks!

stuartyeates:
The simplification is not the point of the commit.

The point of the commit is that if you look at a page like http://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-interaction.html The description for @active includes "Suggested values include: 1] singular; 2] plural; 3] corporate; 4] unknown" but the description for @type lacks this. This text is missing for all tei:valList[@type='closed'].

martindholmes:
I think that since both value lists are included in full below the datatype, a better option might be to remove the list from the other valList types. Why do we need to list those values twice?

stuartyeates:
I'm not sure we do need them twice, one in each of two forms, it's a judgement call, having it inconsistent is just confusing.

If the consensus is to remove, i can come up with another pull request...

martindholmes:
I think we should discuss the ticket. But Council's role with regard to the Stylesheets is very fuzzy indeed.

jamescummings:
Council's role with regard to the Stylesheets is to maintain the stylesheets necessary for the production of the Guidelines and Schemas.  This includes producing the HTML reference pages, chapters, ODD2ODD and ODD2 various schemas.  What it doesn't include (as a requirement) is secondary outputs (e.g. teitoepub, teitoslides) or various import formats (e.g. docxtotei).  Stuart's request involves the output guidelines html pages so very much in the responsibility of the Council.

Reading over the XSL the way github does diffs makes it look like the $firstLang stuff has just vanished. However, looking at the file as a whole it is indeed just further down.  Untested, but a quick skim read looks like this is harmless and clarifying.  I'd either vote for getting rid of them entirely or merging Stuart's pull request. While redundant, it is probably helpful to have the extra text so on balance of probability I'd go for the latter.

-J

stuartyeates:
Maybe I should have opened a sourceforge ticket and linked the two in the comments, then this could be considered like every other thing the council considers?

martindholmes:
No, You're in the right place.

jamescummings:
@stuartyeates Council trying to clean up old PR wonders if this is still a problem... It seems to be fixed already when we look at the website?  Ok to close this PR?

stuartyeates:
This looks fixed to me, thanks @jamescummings

peterstadler:
All seem to be ok with closing this 😄

PR #169, opened 2016-09-01
correct a typo in a LaTeX command

emchateau:
obviously the command subsunsection doesn’t exist in LaTeX

best regards, and thanks a lot for the good work


lb42:
Thanks for noticing the bug!

Issue #230, opened 2017-02-05
JTEI PDF not rendering bibliography as expected


lb42:
Here's my bibliography
~~~~
         <div type="bibliography">
            <listBibl>
               <bibl xml:id="note5">Kent, William <title>Data and reality</title> (North Holland,
                  Amsterdam, 1978)</bibl>
               <bibl xml:id="note6">Wrigley, E.A. <title>Identifying people in the past</title>
                  (E.Arnold, London, 1973)</bibl>
            </listBibl>
         </div>
  ~~~~
It gets rendered like this:

![wotbib](https://cloud.githubusercontent.com/assets/660724/22629790/cd86e3e6-ebe4-11e6-9168-f5b54dda7dc9.png)

Curiously, this is not just prejudice against Mr Wrigley. If I swop the order of the two bibls, it looks like this:

![wotbibl2](https://cloud.githubusercontent.com/assets/660724/22629804/221b267e-ebe5-11e6-9e11-4118e782eb5d.png)

rvdb:
Thanks for notifying. This is happening because a test in a bibliography template expected bibliography entries to contain a tagged authoring instance (&lt;author>, &lt;editor>, &lt;orgName>). If you add them, they will come out as expected.

Yet, your example encoding has made me realize this expectation may have been too limited. I have improved the test, and your encoding now comes out as it should in my local version. Realizing the hardest part is yet to come, I'll postpone the Git commitment struggle for this fix to tomorrow.

lb42:
Thanks for the quick response and fix. I was very puzzled! If the schema had enforced the requirement for an author tag I would have been less puzzled but much more annoyed, so I am  glad it doesn't now need to. But why do I have to supply a div for the bibliography even if I have  nothing to put in it?

PR #231, opened 2017-02-06
made detection of identical authors in consecutive bibliography entri…

rvdb:
…es more robust


rvdb:
fixes https://github.com/TEIC/Stylesheets/issues/230

PR #239, opened 2017-03-11
added checks for trang, ant and xetex to `make check`

bansp:
Modified the Makefile and the readme, to record the information on what extra packages need to be installed. The readme also references Hugh's Docker-related help file.


peterstadler:
While I think that `trang` and `ant`  are probably essential, failing for  `xetex` will exclude all those who simply want to run transformations to Relax and HTML. That should be a warning, no show stopper, IMHO

peterstadler:
Looking at the current repo and build processes, I discovered that we actually do not use locally installed versions of jing and trang (anymore) but those binaries included in the repo. Hence I removed the respective checks.

Issue #261, opened 2017-04-18
TEI address element is wrongly converted to HTML address element


martindholmes:
The purpose of the HTML `<address>` element is very precise and constrained:

~~~~~
The address element provides contact information for a document or part of a document. 
Information provided by address may include the names of the document’s maintainers, 
links to the maintainers’ Web pages, e-mail addresses for feedback, postal addresses, 
phone numbers, and so on. The address element is not appropriate for all postal and 
e-mail addresses; it should be reserved for providing such information about the contact 
people for the document.
~~~~~

The TEI `<address>` element is much more generic, and appears all over the place, containing the addresses of people in personographies, places in gazetteers, and so on. However, the Stylesheets (in particular html/html_core.xsl, line 95) simply converts the TEI `<address>` to the HTML `<address>`, which is both semantically wrong and often invalid (because the range of contexts allowing the HTML address is quite constrained).

We need to decide on another default rendering approach for generic addresses (`<span class="address">`?) for most contexts, and also we need to see if we can figure out when an address _would_ be a good candidate for conversion to the HTML `<address>` element (for instance when it's associated with a document author, a publisher, an editor, or something like that).

lb42:
Ill bet there are plenty of places where html:address is used for its formattibg effect only. And where it really corresponds to something like tei:trailer

jamescummings:
Where can the html:address appear or not? What are the constraints on it?  I would agree that the fall back for tei:address is html:span with a class='address' makes sense since most of the places tei:address is allowed seem to be phrase-level. When you are rendering metadata (e.g. when ancestor::publicationStmt) then perhaps it makes sense to be doing something different with it, but surely that depends on whatever it is we're doing with publicationStmt in our teitohtml at the moment?

martindholmes:
html:address can appear anywhere flow content is allowed (I know that's vague, but I can't find anything more specific). Here's a useful page on it:

https://www.w3.org/wiki/HTML/Elements/address

I think the semantic issues are more significant than the validation issues, although I first came across this through a validation problem in one of our test outputs.

martindholmes:
The Stylesheets co-op group looked at this and decided that in html_core.xsl, the output from `<address>` should be `<span class="address">`, and that css/tei.css should have the following rule added to it:

~~~~~
span.address{
  display: block;
  font-style: italic;
}
~~~~~

That should mean that the output appearance should not change from what it now is, but the invalidity issues should go away. Two other issues came up, which will get separate tickets.

[...]

PR #265, opened 2017-05-04
TEI address element is wrongly converted to HTML address element (#261)

martinascholger:
According to the decision of the Stylesheet co-op group, the output from `<address>` is changed to `<span class="address">` and the CSS selector span.address is added, see the discussion on: https://github.com/TEIC/Stylesheets/issues/261

### DIFF
--- a/css/tei.css
+++ b/css/tei.css
@@ -481,6 +481,11 @@ span.note:after{
 span.note{
     font-size:smaller;
 }
+/* address */
+span.address {
+    display: block;
+    font-style: italic;
+}
 /* images */
 img.display{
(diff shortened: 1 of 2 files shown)


martindholmes:
This looks good to me! Expected fallout will be the usual cascade of errors from expected-results, which someone will have to go through and fix, but perhaps we could attempt to predict and pre-fix them in an another pull request?

peterstadler:
isn't that the chicken and egg problem, that when you change the expected results first, it will choke on those?

martindholmes:
Yes, you'd have to merge both pull requests at the same time. :-)

PR #279, opened 2017-08-25
supplying ns prefix for MEI

peterstadler:
I recently had trouble creating an MEI 3.0 schema with the TEI Stylesheets because it failed to work out the prefix for the MEI namespace needed within the function `tei:generate-nsprefix-schematron`. The following pull request simply supplies this value.

### DIFF
--- a/common/functions.xsl
+++ b/common/functions.xsl
@@ -1308,20 +1308,23 @@ of this software, even if advised of the possibility of such damage.
     <xsl:for-each select="$e">
       <xsl:variable name="myns" select="ancestor::tei:elementSpec/@ns"/>
       <xsl:choose>
-	<xsl:when test="not($myns) or $myns='http://www.tei-c.org/ns/1.0'">
-	  <xsl:text>tei:</xsl:text>
-	</xsl:when>
-	<xsl:otherwise>
-	  <xsl:choose>
-	    <xsl:when test="ancestor::tei:schemaSpec//sch:ns[@uri=$myns]">
-	      <xsl:value-of
-		  select="concat(ancestor::tei:schemaSpec//sch:ns[@uri=$myns]/@prefix,':')"/>
-	    </xsl:when>
-	    <xsl:otherwise>
-	      <xsl:message terminate="yes">schematron rule cannot work out prefix for <xsl:value-of select="ancestor::tei:elementSpec/@ident"/></xsl:message>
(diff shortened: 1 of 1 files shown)


peterstadler:
Sorry, the modification is only lines 1314–1316, the rest is change of indentation.

peterstadler:
Hmm, that doesn't really solve the problem :(

The build succeeds, but the resulting schema now complains about the missing namespace information for the prefix "mei".

For now, I added `<sch:ns uri="http://www.music-encoding.org/ns/mei" prefix="mei"/>` to one of my schematron constraints which seems to remedy the problem. I wonder how the MEI schemas are built?

raffazizzi:
@peterstadler I'm going to have a look, but IIRC I think adding `<sch:ns uri="http://www.music-encoding.org/ns/mei" prefix="mei"/>` is the way to go. I wonder if it got lost in one of the latest updates to the MEI source?

If something's wrong with the TEI XSLTs, then we should look at a generic solution so that ODD can set any namespace, not just TEI's and MEI's.

raffazizzi:
@peterstadler as you found out, the ns is declared in MEI's ODD. 

Strangely, lately the declaration has been moved from the main source to the customization files. I'm not sure why, because I would prefer have the source declare it once rather than burdening a customization writer to provide it each time. But at least it shouldn't matter for the XSLTs.

Example from music-encoding: https://github.com/music-encoding/music-encoding/blob/develop/customizations/mei-CMN.xml#L52
(this is from the development branch, but the master branch has the same decl)

There are still some issues, however, because I'm not getting a correct `<start>` element.  You can see the ns declaration is there, though:
```xml
<start>
    <choice/>
</start>
<ns xmlns="http://purl.oclc.org/dsdl/schematron"
    prefix="mei"
    uri="http://www.music-encoding.org/ns/mei"/> 
```

I think the MEI process is using a pre-PureODD version of the XSLTs to avoid some of these issues that stated popping up. Generally, we (TEI council) should probably have a good chat / call with us (MEI people) to sort a few things out :smiley: 

Closing for now, but we can continue the thread here

peterstadler:
Thanks for your time and sorry for the noise. I am able to create a schema (with the current stylesheets) now and don't face any issue with the start element!

Yet I'm having trouble extracting schematron rules to a separate file … investigating …

PR #285, opened 2017-09-06
avoid double escaping in <egXML> for ampersands that function as enti…

rvdb:
…ty escape character


rvdb:
should address the remaining issues of https://github.com/TEIC/oxygen-tei/issues/6

PR #292, opened 2017-11-02
possible solution for processing listRef/ref in odd2odd.xsl

ebeshero:
I'm assuming that we want to permit something like:
```
               <listRef>
                  <ref target="#NDPER">Some stuff here about processing persNames</ref>
               </listRef>
```
as well as the more usual ptr that resolves to ref in odd2odds:
```
                 <listRef>
                  <ptr target="#NDNA"/>
                </listRef>
```
Since ptr turns into ref in the processing, the ref contents are the head of the given section of the Guidelines in the `@target`. The same thing happens when we permit `listRef/ref`, but what if we want some text the original coder planted in the source ODD? This proposed suggestion provides a way to pull in that text, though I'm not sure that the `<xsl:if>` I'm adding around it is really necessary.


sydb:
We reviewed this yesterday at the Stylesheets co-operative self-educational meeting, and tested it a bit. (Thanks, Elisa!) I see only one tiny tweak I would want to make before merging: the comment on line 2089 should have been moved along with the definition of the variable `$chapter`, i.e. belongs immediately after line 2080.
Problem is, I don’t know how to modify someone else’s pull request. So if @ebeshero makes that change soon, great. If not, I’ll just merge this (into dev, I presume) and then make that tweak after.

ebeshero:
@sydb So moved! :-) That is, I've now moved the comment, and added the change to this pull request.

odds/odd2odd.xsl
```diff
@@ -2071,18 +2071,22 @@ of this software, even if advised of the possibility of such damage.
...
       <xsl:when test="starts-with(@target,'#') and 
-        (ancestor::tei:remarks or ancestor::tei:listRef or ancestor::tei:valDesc) and
+        (ancestor::tei:remarks or parent::tei:listRef or ancestor::tei:valDesc) and
         not(id(substring(@target,2)))">
         <xsl:variable name="target" select="substring(@target,2)"/>
         <xsl:variable name="sourceDoc" select="tei:workOutSource(.)"/>
+        <!-- the chapter ID is on the highest ancestor or self div -->
+        <xsl:variable name="chapter" select="document($sourceDoc)/id($target)/ancestor-or-self::tei:div[not(ancestor::tei:div)]/@xml:id"/>
         <xsl:choose>
+          <xsl:when test="(string-length(normalize-space(.)) &gt; 0) or processing-instruction()">
```
peterstadler: shouldn't we be testing for comment() here as well?

ebeshero: Ah! Good idea--we had not considered this, but I see it's definitely something we should test for...I'll add it to this commit. Thanks @peterstadler

ebeshero:
@peterstadler I've just added the test for `comment()` to this pull request.

ebeshero:
Does anyone spot more possible problems here, or is this ready to pull in?

sydb:
I'm not sure, Peter. Why would we want to copy a comment over? My first instinct (which may be wrong) is that since it's not something that will be processed further down the line is that we would want to drop it, here.

ebeshero:
@sydb I think you weren't following Peter's suggestion (or perhaps I misunderstood it). What I understand is that we need simply to test for the presence of a comment() just as we test for a processing-instruction(), in order to instruct the processing to give us the default behavior that points to the Guidelines.

ebeshero:
@sydb And for what it's worth, we routinely do that elsewhere in this stylesheet, as I quickly saw on review of the code.

sydb:
You may be right, @ebeshero, I may be misunderstanding. But why would we want an output `<ref>` that has no content? The point of the test is to see if there is content to copy over. I am somewhat uncomfortable with the idea that `<ref/>` and `<ref><!-- leave this empty! --></ref>` get processed in different ways.

Yes, the `@target` gets put on the `<ref>`, which probably does not happen in the otherwise clause, but still ...

ebeshero:
@sydb Hmmm...Maybe we did discuss this in that early November call. I'm vaguely remembering us considering the eventuality of comments inside `<ref>` elements, and whether they ought to be carried over into the output...I don't see why they shouldn't?

sydb:
Remember that the point of odd2odd is to merge a main ODD language definition file and an auxiliary ODD customization file into a single “flattened”, “tangled”, “compiled”, or “processed” combined definition of the language. I think the purpose of that compiled output is for machine consumption. That is, except for debugging, no one should be reading it. It is deliberately intended to be the input to further processing. As evidence, note that the output of odd2odd does not contain any of the chapter prose.
Thus the idea that it is important to retain comments, which exist only for humans, not for computers, is an anathema (is that the right word?). One could argue “but if the comment isn’t doing any harm, why not keep it?”. My answers are that the same is true of the chapter prose, which we don’t keep; and that if the processing of `<ref/>` and `<ref><!-- --></ref>` is different, then keeping it is doing harm (to my thought process, if nothing else :-).

peterstadler:
wait @sydb, I thought you were listening to my talk in Victoria ;) 

>  It is deliberately intended to be the input to further processing. As evidence, note that the output of odd2odd does not contain any of the chapter prose.

Yes, odd2odd is the beginning of it all, including documentary output formats such as HTML and PDF. And it's not odd2odd that's throwing away the chapter prose but we are not including it in the source p5subset.xml ! This is some other thing!

> Thus the idea that it is important to retain comments, which exist only for humans, not for computers, is an anathema

Well, I wouldn't say that the “flattened”, “tangled”, “compiled”, or “processed” ODD should be read by a human *but* it should carry on as much information as possible to the next step in the processing pipeline—which might be read by a human. So, I'd argue strongly for keeping those comments …

ebeshero:
@peterstadler I *was* listening to your talk, and I have been thinking a lot about how we use ODDs in projects lately with students and newcomers who are just learning TEI. We are not exactly in danger of losing official Guidelines documentation, which everyone can readily locate. What doesn’t make sense is dropping our localized documentation just because the next stage of machine processing doesn’t need it. That sounds good in the name of efficiency and not making @sydb ‘s head hurt, but maybe we ought to be helping our community to preserve its comments as documentation. A comment inside an element isn’t equivalent intellectually to an empty element, after all.

sydb:
@peterstadler: you talked?

Seriously, I am not completely convinced either way on this one,[1] but just to be clear —

> A comment inside an element isn’t equivalent intellectually to an empty element, after all.

Yes, actually, they may be precisely equivalent. Per the XML Spec section 2.5 “Comments”:

> [Comments] are not part of the document's character data; an XML processor … need not make it possible for an application to retrieve the text of comments.

That is to say any step along the way is welcome to throw away comments. Of course a comment is part of the XDM, and we do have access to them via XPath — XSLT is an XML process that happens to keep comments (explicitly, I think, but I cannot get access to the spec now). But relying on them is walking on thin ice.

[1] The main reason I’m not convinced is that it makes so little difference: the only difference whether `<ref><!-- duck --></ref>` is captured by that first `<xsl:when>` or by the `<xsl:otherwise>` is whether or not the `@target` and comment end up in the output or not. I am strongly suspicious that whichever way we do it, someone at some point will be unhappy with it.

ebeshero:
@sydb Can we look at other places in  odd2odd.xsl where we process comment()? When I did so, I saw a pattern of copying these over alongside processing instructions and text nodes, which is why @peterstadler request seemed consistent with practices elsewhere on the stylesheet. There is just one place where I'm noticing suppression of comments, on line 743, and I think that probably isn't a universal suppression given that they are frequently processed by xsl:copy-of in other template rules.

ebeshero:
I've just updated my fork, so this pull request is up-to-date with the current state of Stylesheets dev.

PR #295, opened 2017-11-15
Add makeAnchor to tei:ab where it generates html:div

TomazErjavec:
In the digital edition http://nl.ijs.si/e-zrc/kapelski/index-en.html we now make reference (via ref/@target) to ab/@xml:id, which didn't work in HTML because tei:ab/@xml:id didn't get mapped to html:div/@id. And this is now fixed.
Btw., this is my first pull request, so I hope it is kosher!


peterstadler:
seems reasonable,  tested and merged it.

Many thanks @TomazErjavec !

Issue #299, opened 2017-11-30
docx2tei conversion produces incorrect ref/@target values


rvdb:
In the docx2tei conversion, the processing of `<w:hyperlink>` produces incorrect `tei:ref/@target` attribute values:

- whenever a docx hyperlink contains a fragment identifier in `w:hyperlink/@w:anchor`, only the fragment identifier is copied to the resulting `tei:ref/@target` attribute, while the URL is discarded. For example: 

      <w:hyperlink r:id="rId11" w:anchor="SAPT" w:history="1"></w:hyperlink>

  ...inside `document.xml` with the target URL encoded in a corresponding entry in a separate file `document.xml.rels`:

      <Relationship Id="rId11" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/hyperlink" Target="http://www.tei-c.org/release/doc/tei-p5-doc/en/html/SA.html" TargetMode="External"/>

  When converting this to TEI, this is processed to:

      <ref target="SAPT"></ref>

  instead of:

      <ref target="http://www.tei-c.org/release/doc/tei-p5-doc/en/html/SA.html#SAPT"></ref>
-  for hyperlinks without fragment identifier that occur inside endnotes, the target URL in the generated `tei:ref/@target` attribute is pointing to (various) docx helper files, instead of the original URL. For example:

       <w:hyperlink r:id="rId1" w:history="1"></w:hyperlink>

   ...inside `endnotes.xml`, with the target URL encoded in a corresponding entry in a separate file `document.xml.rels`:

       <Relationship Id="rId1" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/hyperlink" Target="http://www.tei-c.org/release/doc/tei-p5-doc/en/html/SA.html" TargetMode="External"/>

   When converting this to TEI, this is processed to:

       <ref target="../customXml/item1.xml"></ref>

   instead of:

       <ref target="http://www.tei-c.org/release/doc/tei-p5-doc/en/html/SA.html"></ref>

I'll attach a zip file [hyperlinktest.zip](https://github.com/TEIC/Stylesheets/files/1517896/hyperlinktest.zip), containing:

- hyperlinktest.docx: a sample docx file
- hyperlinktest.xml: a resulting TEI file, processed using the latest stylesheets code (be2a2e8)

I'm preparing a pull request that solves this issue.

PR #300, opened 2017-11-30
docx2tei hyperlink fix

rvdb:
fixes https://github.com/TEIC/Stylesheets/issues/299


peterstadler:
Hi @rvdb, I took a quick a look at the changes but it was not self-explanatory to me. Could you share some background for this PR and a test case/example? Many thanks!

rvdb:
Hi @peterstadler , sorry for the brevity. I hope the files attached will clarfiy the fix. I've found two issues which I've addressed in two commits:
  - 8d91eb3 : the current stylesheets produce incorrect `tei:ref/@target` values for hyperlinks containing fragment identifiers: instead of joining the URL and fragment identifier, only the fragment identifier is output. For example: a link to http://www.tei-c.org/release/doc/tei-p5-doc/en/html/SA.html#SAPT in a DOCX file is being transformed to `<tei:ref target="SAPT">...</tei:ref>` instead of `<tei:ref target="http://www.tei-c.org/release/doc/tei-p5-doc/en/html/SA.html#SAPT">...</tei:ref>`. This happens because the fragment identifier check is too greedy, which I've fixed in this commit.
  - d539356 : for hyperlinks in endnotes, apparently the wrong processing context is passed, which results in incorrect lookups for the URL part of a link. A link to http://www.tei-c.org/release/doc/tei-p5-doc/en/html/SA.html inside an endnote is being transformed to `<tei:ref target="../customXml/item1.xml">...</tei:ref>`, instead of `<tei:ref target="http://www.tei-c.org/release/doc/tei-p5-doc/en/html/SA.html">...</tei:ref>`. This happens because the endnote context node itself is not copied properly in the conversion chain (whereas footnotes were copied correctly). When hyperlinks inside endnotes are processed further down the line, the missing endnote context caused incorrect lookups of helper files, and hence incorrect `tei:ref/@target` values in the generated output. In this commit, I've made sure the endnote context is copied properly, in analogy to how footnotes are being processed.

If all goes well, you'll find attached a [hyperlinktest.zip](https://github.com/TEIC/Stylesheets/files/1651346/hyperlinktest.zip) file, containing:
  - hyperlinktest.docx: a test DOCX file
  - hyperlinktest-orig.xml: TEI output of the original stylesheets
  - hyperlinktest-8d91eb3.xml: TEI output after commit 8d91eb3, fixing the fragment identifier issue
  - hyperlinktest-8d91eb3+d539356.xml: TEI output after commits 8d91eb3 and d539356, fixing the improper processing context for hyperlinks inside endnotes as well

rvdb:
@peterstadler I think the PR provides a limited fix to a well isolated problem, and I hope I've clarified what it does, and why. Do you need any more information/details?

peterstadler:
Sorry for the delay; it may look simple to you but it's much harder for me ;)

So, I checked out your fix and ran `make  test-from-docx` (from within the Test subfolder) which gave an error regarding `expected-results/test39.xml`. The diff looks like: 

```
<         <ref target="#_Toc382049195">Introduction	2</ref>
---
>         <ref target="#Toc382049195">Introduction	2</ref>
60c60
<         <ref target="#_Toc382049196">Using standard Word formatting	2</ref>
---
>         <ref target="#Toc382049196">Using standard Word formatting	2</ref>
…
```

It looks as if your commit (re)introduced the underscores to the fragment identifiers. The whole thing is a little bit odd, because the underscores stem from the original docx file and magically are removed from the TEI result. Magically, because I could only find https://github.com/TEIC/Stylesheets/blob/c1b8f67fbbd350f1ffaa309c30e968796caacb15/docx/from/pass2.xsl#L404 where the underscore of the target `@xml:id` gets removed.

That said, I have some sympathy for keeping the underscores (= your version), BUT:

* someone must have introduced that feature on purpose. What was it?
* I do not see any bit in the old code that removes that underscore, but there must be. So, are there other areas affected?
* if we go with the underscores, we'd need to fix the targets as well
* finally, we'd need to update the expected results (with this PR)

I'll invite @lb42 and @jamescummings to join the discussion, maybe they have a clue about the historic idea of removing those underscores?

rvdb:
@peterstadler Thanks for investigating, and I didn't want to underestimate the complexity of the stylesheets, nor your involvement with them!

I see the difference, and -thanks to your pointer- I've been able to locate its origin in `pass2` mode, where `<tei:ref>` elements are further processed at: https://github.com/TEIC/Stylesheets/blob/c1b8f67fbbd350f1ffaa309c30e968796caacb15/docx/from/pass2.xsl#L587 The underscores are being removed from the `@target` attributes in the `tei:processInstruction()` function: https://github.com/TEIC/Stylesheets/blob/c1b8f67fbbd350f1ffaa309c30e968796caacb15/docx/from/functions.xsl#L224 
There you'll find a lot of regex matching for `'^_'` patterns; for normal hyperlinks as this one, only the one at L246 is at play: https://github.com/TEIC/Stylesheets/blob/c1b8f67fbbd350f1ffaa309c30e968796caacb15/docx/from/functions.xsl#L246
The original code in: https://github.com/TEIC/Stylesheets/blob/c1b8f67fbbd350f1ffaa309c30e968796caacb15/docx/from/docxtotei.xsl#L519 just passes the "bare" anchors, but since this was the problem I've tried to correct in my PR, I've changed this so this template always composes "full" targets, either consisting of the target URL (e.g. http://www.tei-c.org/release/doc/tei-p5-doc/en/html/CO.html), only a fragment identifier in the Word file (e.g. #localtarget), or both (e.g. http://www.tei-c.org/release/doc/tei-p5-doc/en/html/CO.html#COXR). 

Hence, with my PR, the `tei:processInstruction()` function won't see just the "bare" fragment identification codes, but instead, the `@target` attributes it is processing will always be preceded by either the target URL or a '#'. Therefore, changing the code at L246: https://github.com/TEIC/Stylesheets/blob/c1b8f67fbbd350f1ffaa309c30e968796caacb15/docx/from/functions.xsl#L246
to this:

    <xsl:value-of select="if (matches($instr,'(^|#)_')) then replace($instr, '(^|#)_', '$1') else $instr"/>

...will remove the underscores back again and fix the failing test.

If it could help, I could make a different PR for this, but as I don't know how to include this into this PR in Git, I'll let you decide what is the better option.

peterstadler:
In a hurry: Please add your changes to your [docx-hyperlink-fix](https://github.com/rvdb/Stylesheets/tree/docx-hyperlink-fix) branch and this PR should automagically be updated.

rvdb:
@peterstadler Ok, thanks. I've briefly tried to run the `make` command from a Cygwin bash, but that didn't work. Anyway, I've tested the individual transformation of `test39.docx` and compared the output to `expected-results/test39.xml`, and the link targets are identical now.

peterstadler:
All Stylesheets tests have passed. Thanks @rvdb for the very precise bug description (as always) and remedy!

PR #327, opened 2018-06-20
use ODT 'Author' style to specify author

petrul:
this change allows the specification of the Author of the document also by using the 'Author' ODT Style (alternatively to the hard-to-access dc:initial-creator meta tag).

The 'Title' ODT Style is already used to allow the specification of the title of the document (alternatively to dc:title). This change extends the same logic for the Author.

(in libre office meta fields are either hard to access or downright missing in the GUI).


peterstadler:
I fear something went wrong with this PR as there are no changes to the code? Or am I missing something?

odt/odttotei.xsl
```diff
@@ -125,6 +125,23 @@ of this software, even if advised of the possibility of such damage.
...
 
+  <xsl:variable name="document-author">
+    <xsl:choose>
+      <xsl:when
+	  test="/office:document-content/office:body/office:text/text:p[@text:style-name='Author']">
+        <xsl:value-of
+	    select="/office:document-content/office:body/office:text/text:p[@text:style-name='Author']"
+	    />
+      </xsl:when>
+      <xsl:when test="$META/office:meta/meta:initial-creator">
```
peterstadler: To reproduce the old version you should probably only take the first `text:p[@text:style-name='Title']` (see old line 116), hence this should be `$META[1]/office:meta/meta:initial-creator`

odt/odttotei.xsl
```diff
@@ -125,6 +125,23 @@ of this software, even if advised of the possibility of such damage.
     </xsl:choose>
   </xsl:variable>
 
+  <xsl:variable name="document-author">
+    <xsl:choose>
+      <xsl:when
+	  test="/office:document-content/office:body/office:text/text:p[@text:style-name='Author']">
+        <xsl:value-of
+	    select="/office:document-content/office:body/office:text/text:p[@text:style-name='Author']"
```
peterstadler: I'm not familiar with the odt format but I guess there might be several authors (see also next comment). So it's probably safest to pick only the first `text:p[@text:style-name='Author']`. Or we need to iterate over the authors when outputting these …

peterstadler (commented):
I tested locally without any errors! That's the good news ;)
The bad news is, it seems there is some dependent issue with multiple authors? I'm not familiar with the odt format, but my guess it we should be able to support this (or restrict to one)?

odt/odttotei.xsl
petrul: Indeed the old version only picked the first 'Title'-styled paragraph. 

I see no reason for that. If the Title is streched on two lines, both should be taken.

Same rationale for the Author, especially given that one document might have several Authors

peterstadler: Yes, but within the resulting TEI document there should be an `<author>` element for every author name (not several names within one `<author>` element).
Could you add (and push) the appropriate `<xsl:for-each>` loop to your branch (so the pull request will be updated)?

petrul:
Hello again, sorry for the delay, I had completely forgotten XSLT.

As I had to dive again recently into this marvellous set of tags language, I could also fix your request to generate several <author> </author> elements inside the titleStmt, if there are several paragraphs in the .odt styled with 'Author'.

There are also a few additions, mainly taking care of special styles: Subtitle, Epigraph, Trailer etc which generate <p class="subtitle"> HTML elements.

martindholmes:
@peterstadler to go ahead, but look carefully at each individual change; for example, `<p class="subtitle">` seems a bit strange to me, unless I'm misunderstanding it.

martinascholger:
The Stylesheets group will have a look at it in the next meeting in October

martindholmes:
@sydb, @npcole and I looked at this today, and we believe that the original pull request has been severely muddied by additional commits to the same PR. It's also not clear whether @petrul did address @peterstadler's concerns about multiple authors, or the additional concern raised about multi-line titles. We suggest that each of these specific issues be raised as individual tickets and implemented.

odt/odttotei.xsl
```diff
@@ -877,7 +877,7 @@ of this software, even if advised of the possibility of such damage.
   </xsl:template>
 
   <xsl:template match="office:binary-data">
-    <binaryObject encoding="base64" mimeType="image/jpg">
+    <binaryObject id="{generate-id()}" encoding="base64" mimeType="image/jpg">
```
sydb: @petrul — TEI does not have an `@id` attribute. You mean `@xml:id`, yes?

sydb (changes requested):

petrul:
Indeed, it would be the regular xml id attribute.

odt/odttotei.xsl
petrul: for me it s a mystery why this commit ended up on this merge request, I just modified my own fork of tei-stylesheets and commited. I guess there's more to git than you can learn in a few years.

It these changes do not make sense for you, feel free to disregard this merge request.

raffazizzi:
We couldn't merge this PR, but we plan to implement this later on, see https://github.com/TEIC/Stylesheets/issues/625

PR #332, opened 2018-08-29
Implementation of sorted diffing to make Test more robust against software version changes.

martindholmes:
Recent version of debian-based OSes have software changes which result in the sequence of items in a zip file listing being in a different order from those in older versions. When we do diffs of file listings against expected-results in the Stylesheets Tests, those tests fail because the expected-results are in a different order. The sequence of items is insignificant, so sorting both the test file and the expected-results file before diffing results in an equally rigorous test but makes it more robust against software changes, allowing the tests to succeed on (for example) Ubuntu 18.04, the current LTS.


hcayless:
I think this is the way to go.

PR #336, opened 2018-09-13
adding rendering of ref, gi and att inside desc

PietroLiuzzo:
I wanted the Relax-NG output from my ODD to render ref[@target] as link and gi | att as code, adding the three small templates does it for this purpose and gets visualized in oXygen, but I ignore other possible implications. I have tested modifying my local version of the stylesheets.

### DIFF
--- a/odds/teiodds.xsl
+++ b/odds/teiodds.xsl
@@ -2456,6 +2456,17 @@ select="$makeDecls"/></xsl:message>
   </xsl:template>
 
 
+  <xsl:template match="tei:ref[@target]">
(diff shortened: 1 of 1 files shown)


sydb:
Council sub-group just tested, and this PR seems fine. There are some outstanding issues:

1. The expected results have not been changed, thus a build will fail until we fix that. We don't want to fix that on the PR, because there have been other changes that probably effect the expected results since. So we'll accept this and updated expected results later (probably this afternoon).
1. We realize that the newly inserted code here does not process `<gi>`s the same way the Guidelines do, and perhaps should.
1. We don't yet know if there are already templates within the Stylesheets that should have been used instead, but were not firing for some reason.

PietroLiuzzo:
thanks!

PR #339, opened 2018-10-22
Make editorial expansions appear as italic in HTML

cyocum:
Make editorial expansions appear as italic in HTML. This allows
editorial interventions to be visible when a transcription is
translated into HTML.  This is common formating for printed critical
editions.


hcayless:
My instinct is to say “no” to this. This should be done with CSS or with a local customization. In my discipline, `<ex>` would be wrapped in parentheses. Why are italics better?

peterstadler:
but `<span class="ex">`, instead of `<i>`, would be an improvement, wouldn't it?

cyocum:
> Looking at the failed test, the result has an additional `<i>` within the `<span class="corr">[gvn<i>n</i>hilldr]</span>`. Is this expected? Then could you please add it to `Test/expected-results/test.html`.
> 

No, sadly that was an oversight on my part. I will fix it presently.

> My instinct is to say “no” to this. This should be done with CSS or with a local customization. In my discipline, `<ex>` would be wrapped in parentheses. Why are italics better?

Italics are what is used in my discipline (Celtic Studies/Early Medieval Irish) when we expand scribal contractions in MSS when doing a diplomatic edition.  This is the convention but I do not feel one is better than the other as long as it is marked in some way.  My users will expect italics but as long as the expansions are marked in some way, that will be acceptable.

> but `<span class="ex">`, instead of `<i>`, would be an improvement, wouldn't it?

I am fine with using the above. I can withdraw the PR and resubmit once we have figured out what (italics or parenthesis) we want to use.

I did this while on holiday and I will be boarding a plane soon so any amendments or re-submission will probably need to wait until the weekend or early next week.

jamescummings:
Agreed with the above -- although in anything I edit expanded text is shown in italics, it is not standard in all disciplines. (While I find classicists love of parentheses unreadable, I understand it dates from a time before easy italics ;-) ) 

Generally when converting from TEI to HTML previously I think we've always preferred `<span class="elementName">` though you could argue that to avoid any name collisions we could be moving to something like `<span class="tei_elementName">` e.g. `<span class="tei_ex">` for expanded text. But if changing that we should standardise across the various HTML output of classes names.

martindholmes:
Agreed -- definitely avoid HTML elements which are pure style (i, u, b) in favour of either semantic elements (emph etc.) and/or classes. I like to use the source TEI element names where possible, but that's often not enough when it's a generic element.

cyocum:
I have made a change where I use `<span class="ex">` and I have added CSS to the tei.css file.  Hopefully, this will be acceptable. I am not sure why the tests are failing now and I cannot seem to rerun them from here but the code is outputting what I would expect.  Thank you for your input on this.

peterstadler:
I'm about to update the remaining tests …

cyocum:
If you would like to do that, it would be very much appreciated :).  I am blocked up at my day job for the next couple of days so I would not have a chance to do it until then.

peterstadler:
ok, tests have passed. May I ask @jamescummings to review again and merge if possible :)

jamescummings:
Changes on pull request reviewed, merged in. Thanks all for the changes!

Issue #358, opened 2019-01-21
OxGarage Does Not Preserve Manual Page Breaks in TEI P5


mysheepb:
It seems that OxGarage does not keep the manual page breaks in DOCX or ODT documents when converting to TEI documents. If I want to keep the manual page breaks in TEI documents, using the <pb /> tag, what should I do? Can I modify a "profile" or code for this purpose?

peterstadler:
I took a look at the Stylesheet `odttotei.xsl` and think I can up with a solution. PR coming …

mysheepb:
That's great! Thank you so much for your help!

peterstadler:
I just created the PR #359. Feel free to try it out—Feedback much appreciated!

Anyway, thanks for spotting and notifying!

mysheepb:
I have tried the fix and it works well! Thanks a lot!

PR #359, opened 2019-01-24
add support for hard page breaks in odttotei

peterstadler:
This PR addresses #358 and adds `<tei:pb>` (when indicated by odt styles) for
* paragraphs `<text:p>`
* tables `<table:table>`
* lists; admittedly not very elegant, because the page break is actually inserted via the `<text:p>` of the first list item. Thus, the `<tei:pb>` ends up _within_ the `<tei:list>`

### DIFF
--- a/Test/expected-results/test7.xml
+++ b/Test/expected-results/test7.xml
@@ -128,6 +128,7 @@
                </list>
             </div>
          </div>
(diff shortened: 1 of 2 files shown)


peterstadler:
pinging @jamescummings for review …

Issue #357, opened 2019-01-18
Extra space before semicolon in list of suggested values in the RELAX NG


sydb:
In the annotations that list the (suggested|sample) values include in the RELAX NG there is an extra space before the separating semicolon after a value iff there is a `<gloss>` for that value.

sydb:
@martindholmes and I have found the source of the problem, which is the `tei:makeGloss()` function in common/functions.xsl, as called from line 1199.

peterstadler:
is that related or maybe introduce by #137 ?
(just for reference of the issue)

martindholmes:
I'm pretty sure it wasn't introduced by #137's fix, because it already appears in the quoted text which forms part of the bug report in #137. :-)

sydb:
Stylesheets group discussed this. We think the best idea is to change tei:makeGloss() so that it returns a single string (rather than node()*), which has no leading or trailing space; and change the 6 places where tei:makeGloss() is called so each expects that (a single string w/o leading or trailing whitespace). We also think this is the hardest fix to implement.

Easier would be to change the one place where tei:makeGloss() is called that causes **this** problem. (Roughly line 1198 of common/functions.xsl.) Which is what we ended up doing. Note that we had to use `<value-of>` rather than `<sequence>`, as the latter serializes the items in the sequence with a space between ’em.

martindholmes:
I think the merge of Peter's pull request to fix the broken build has undone this fix. Needs to be re-implemented. I'm not sure why this happened.

peterstadler:
That's hopefully not the case though I admit that I tweaked the processing a little bit in 
f4e343d (but referenced it here properly ;) 
The rest of the change from 7f30618 is untouched, see https://github.com/TEIC/Stylesheets/blob/f4e343dbb264d4ab968e3c7775fcd720018fa592/common/functions.xsl#L1201

peterstadler:
@martindholmes can you give an example where you see that extra space before the separating semicolon?

martindholmes:
Sorry @peterstadler, my mistake. My Test2 expecte-results were out of date.

PR #364, opened 2019-03-07
set localsource parameter once for build process 

peterstadler:
This PR fixes the current build process issue due to http->http redirects for http://www.tei-c.org/Vault/P5/current/xml/tei/odd/p5subset.xml. 

Rather than changing all calls in the respective stylesheets I tried to improve the build process by adding and propagating a `DEFAULTSOURCE` parameter, so finally it's possible to run the tests against a local source of your choice! 

There were several changes needed:

* `tools/getnames.xsl`: replace hard coded value with a parameter that can be overwritten from outside
* `bin/transformtei`: support remote resource for `--localsource` parameter
* `Makefile` and `Test/Makefile`: add defaultSource parameter per default to every call to saxon, and add the `--localsource` parameter per default to every call to `/$(BINDIR)/teitoXXX` 
* some updates to the expected results as aftermath of #357 
* and well, this does not belong here, but when diffing the tests with the expected results I found an extra trailing whitespace from #357 which is removed now in `common/functions.xsl`

### DIFF
--- a/Test/Makefile
+++ b/Test/Makefile
@@ -3,7 +3,7 @@ VERSION=`cat VERSION`
 REGENERATING=0
 # 0 = do diff; 1 = dont do diff
 BINDIR=../bin
-DEFAULTSOURCE=https://www.tei-c.org/release/xml/tei/odd/p5subset.xml
+DEFAULTSOURCE=https://www.tei-c.org/Vault/P5/current/xml/tei/odd/p5subset.xml
 SAXONOPT=useFixedDate=true
 SAXON=java -Djava.awt.headless=true -jar ../lib/saxon9he.jar defaultSource=$(DEFAULTSOURCE)
 DOTDOTSAXON=java -Djava.awt.headless=true -jar ../../lib/saxon9he.jar defaultSource=$(DEFAULTSOURCE)
--- a/Test/Makefile
+++ b/Test/Makefile
@@ -1,13 +1,15 @@
 #added REGERATION option (LB 2015-12-16)
 VERSION=`cat VERSION`
-FLAG=
 REGENERATING=0
 # 0 = do diff; 1 = dont do diff
 BINDIR=../bin
+DEFAULTSOURCE=https://www.tei-c.org/release/xml/tei/odd/p5subset.xml
 SAXONOPT=useFixedDate=true
-SAXON=java -Djava.awt.headless=true -jar ../lib/saxon9he.jar
-DOTDOTSAXON=java -Djava.awt.headless=true -jar ../../lib/saxon9he.jar
-JING=java -Djava.awt.headless=true -jar ../lib/lib/jing-20120724.0.0.jar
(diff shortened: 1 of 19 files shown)


martindholmes:
This looks good to me.

peterstadler:
> Should the warning be updated to https?
> 
> `echo "WARNING: No localsource set. Will get a copy from http://www.tei-c.org/Vault/P5/."`

Maybe even refrain from outputting a 'real' URL but simply state "from the TEI Vault" or similar?

martindholmes:
> > Should the warning be updated to https?
> > `echo "WARNING: No localsource set. Will get a copy from http://www.tei-c.org/Vault/P5/."`
> 
> Maybe even refrain from outputting a 'real' URL but simply state "from the TEI Vault" or similar?

I think the real URL should be shown, to help people debug problems, but it should presumably come from a variable rather than be a literal.

Test/Makefile
```diff
@@ -390,6 +390,7 @@ test-oddity: css
 	$(SAXON) -s:testdrama.odd -xsl:../odds/odd2odd.xsl -o:testdrama.compiled.xml
 	perl -p -i -e 's+ xml:base=".*testdrama.odd"++' testdrama.compiled.xml
 	perl -i -pe 'BEGIN{undef $$/;} s/<\?TEIVERSION[^?]+\?>//smg' testdrama.compiled.xml
+	perl -i -pe 's+(schemaSpec.*source=")$(DEFAULTSOURCE)+$$1https://www.tei-c.org/Vault/P5/current/xml/tei/odd/p5subset.xml+' testdrama.compiled.xml
```
martindholmes: Suggest adding a comment like this: 

If the compiled odd is created with a default source that is different from the main vault current p5subset.xml, the diff against expected-results will fail, so we just fix it to avoid that failure.

It took Syd and I a little while to figure out exactly what this was doing.

bin/transformtei
```diff
@@ -166,16 +166,35 @@ then
...
 if test "x$TEISOURCE" = "x"
 then
   localsource=""
-  echo "WARNING: No localsource set. Will get a copy from http://www.tei-c.org/Vault/P5/."
-else 
-#  REALD=`dirname "$TEISOURCE"`
-#  SDIR=`(cd $REALD; pwd)`
-#  REALSOURCE=$SDIR/`basename $TEISOURCE`
-  echo using $TEISOURCE as default source
+  echo "WARNING: No localsource set. Will get a copy from http://www.tei-c.org/Vault/P5/ when necessary."
```
sydb: @martindholmes and I think this echo command should have the URL as a variable, so it is set correctly.

sydb:
@martindholmes and I have looked through this PR and tested it locally, and it doesn't seem to make anything worse. Given that the build is broken, and this fixes that, we think it is a DYNAMITE change and are applying it presto. There are a few places where tiny changes (to echo cmds or to comments) are in order, but they are not enough to give us pause merging this.

PR #367, opened 2019-03-22
make sure to only pass one prefix to `concat()`, fixes 366

peterstadler:
that originated from https://github.com/TEIC/Roma/issues/26

While it looks like a simple fix, I'm not 150% sure about the implications, so feedback from @sydb who's done more work on Schematron is highly appreciated.

Also, @cmsmcq has a local fix for the issue at hand. Maybe he can add to that as well?

### DIFF
--- a/common/functions.xsl
+++ b/common/functions.xsl
@@ -1311,7 +1311,7 @@ of this software, even if advised of the possibility of such damage.
             * Note that $myns will be undefined both when the ancestor
             * <elementSpec> does not have an @ns, and when there is no
             * ancestor <elementSpec>. *-->
-        <xsl:when test="not($myns)  or  $myns='http://www.tei-c.org/ns/1.0'">
+        <xsl:when test="not($myns)  or  $myns eq 'http://www.tei-c.org/ns/1.0'">
           <xsl:text>tei:</xsl:text>
         </xsl:when>
         <!--* otherwise, if an ancestor has a suitable <sch:ns> element, 
--- a/common/functions.xsl
+++ b/common/functions.xsl
@@ -1298,29 +1298,56 @@ of this software, even if advised of the possibility of such damage.
   <doc xmlns="http://www.oxygenxml.com/ns/doc/xsl">
     <desc>which prefix for schematron</desc>
   </doc>
-
+  <!--* tei:generate-nsprefix-schematron($e)
+      * Calculate a namespace prefix for a given element, or fail. *-->
   <xsl:function name="tei:generate-nsprefix-schematron" as="xs:string">
-    <xsl:param name="e"/>
+    <xsl:param name="e" as="element()"/>
     <xsl:for-each select="$e">
-      <xsl:variable name="myns" select="ancestor::tei:elementSpec/@ns"/>
+      <xsl:variable name="myns" select="normalize-space( ancestor::tei:elementSpec/@ns )"/>
+      <xsl:variable name="prefixes" select="in-scope-prefixes(.)" as="xs:string*"/>
+      <xsl:variable name="uris" select="for $pre in $prefixes return namespace-uri-for-prefix( $pre,$e)" as="xs:anyURI*"/>
       <xsl:choose>
-        <xsl:when test="not($myns) or $myns='http://www.tei-c.org/ns/1.0'">
+        <!--* if ns is not specified or is tei, use 'tei'
+            * Note that $myns will be undefined both when the ancestor
+            * <elementSpec> does not have an @ns, and when there is no
+            * ancestor <elementSpec>. *-->
+        <xsl:when test="not($myns)  or  $myns='http://www.tei-c.org/ns/1.0'">
           <xsl:text>tei:</xsl:text>
         </xsl:when>
+        <!--* otherwise, if an ancestor has a suitable <sch:ns> element, 
+            * use the one belonging to the nearest such ancestor *-->
+        <xsl:when test="ancestor::*/sch:ns[ normalize-space(@uri) eq $myns ]">
+          <xsl:variable name="a" as="element()"
+                        select="ancestor::*[sch:ns[ normalize-space(@uri) eq $myns ]][1]"/>
+          <xsl:variable name="prefix" as="xs:string*"
+                        select="$a/sch:ns[ normalize-space(@uri) eq $myns ]/normalize-space(@prefix)"/>
+          <xsl:value-of select="concat( $prefix[1],':')"/>
+        </xsl:when>
+        <!--* Otherwise, if the actual declaration has a suitable
+            * namespace node, use the current prefix. Or, to be
+            * exact, any one of the current non-null prefixes.
+            * We'll take the first one presented.
+            *-->
+        <xsl:when test="$uris = $myns  and  $myns ne ''">
+          <xsl:variable name="indx" select="index-of( $uris, $myns )[1]"/>
+          <xsl:value-of select="concat( $prefixes[$indx],':')"/>
+        </xsl:when>
+        <!--* If no ancestor has a suitable sch:ns child, and we don't
+            * have any local namespace bindings with non-zero
+            * prefixes, then we are desperate and we will take any 
+            * sch:ns in the schemaSpec (this is getting a bit desperate) *-->   
+        <xsl:when test="ancestor::tei:schemaSpec//sch:ns[ normalize-space(@uri) eq $myns ]">
+          <xsl:variable name="NSs" select="ancestor::tei:schemaSpec//sch:ns[ normalize-space(@uri) eq $myns ]"/>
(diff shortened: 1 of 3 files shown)


martindholmes:
This assumes that there is definitely at least one sch:ns declaration. Are there scenarios in which there might be none?

cmsmcq:
@peterstadler hints that he would like to see my local fix for the odd2odd.xsl stylesheet.  I append it.  (My apologies for  not presenting it as a pull request; I am out of practice and just giving it here seems likely to be much faster.)  Like the pull request on which this is a comment, my fix essentially ensures that only one prefix is returned; it differs from pull request 367, however, in a couple of ways.  At the risk of providing too much information, I'll go through the entire `generate-prefix-schematron` function.

The general structure of the function and the first few lines are functionally the same as in the current code; the only difference is that I have added comments to guide the maintenance programmer.

    
    <xsl:function name="tei:generate-nsprefix-schematron" as="xs:string">
      <xsl:param name="e" as="node()"/>
      <xsl:for-each select="$e">
        <xsl:variable name="myns" select="ancestor::tei:elementSpec/@ns"/>
        <xsl:choose>

The first case deals with the possibility that `tei:elementSpec/@ns` is not specified or is the TEI namespace; it's the same as the current code.

          
          <xsl:when test="not($myns) or $myns='http://www.tei-c.org/ns/1.0'">
            <xsl:text>tei:</xsl:text>
          </xsl:when>

The next case is the one that changes.  (The current code has a pointless nested xsl:choose, which I have flattened.)

The current code looks for all matching `sch:ns` elements which are descendants of any enclosing `tei:schemaSpec` element, takes the `@prefix` attribute, and returns them all.

            
            <xsl:when test="ancestor::tei:schemaSpec//sch:ns[@uri=$myns]">
              <xsl:value-of
                  select="concat(ancestor::tei:schemaSpec//sch:ns[@uri=$myns]/@prefix,':')"/>
            </xsl:when>

(The condition on the `xsl:when` incidentally also provides the first part of the answer to the question asked by @martindholmes: if there is no matching `sch:ns` element, we skip this case and move to the next.)

Peter Stadler's proposed change changes the code by taking the first such prefix.  That will work for any Odd file in which the given namespace is bound to at most one prefix, because then the sequence of prefixes found will all be the same.

            
            <xsl:when test="ancestor::tei:schemaSpec//sch:ns[@uri=$myns]">
              <xsl:value-of
                  select="concat((ancestor::tei:schemaSpec//sch:ns[@uri=$myns]/@prefix[1]),':')"/>
            </xsl:when>

The code in my local fix tries to be prepared for the possibility that the prefixes might not always be the same, for whatever reason, so it takes not the first prefix in the document but the one given by the nearest ancestor which provides a binding.
	  
          
          <xsl:when test="ancestor::*/sch:ns[@uri=$myns]">
            <xsl:variable name="a" as="element()"
                          select="ancestor::*[sch:ns[@uri=$myns]][1]"/>
            <xsl:variable name="prefix" as="xs:string"
                          select="$a/sch:ns[@uri=$myns]/@prefix"/>
            <xsl:value-of select="concat($prefix,':')"/>
          </xsl:when>

I note in passing that when I wrote this, I seem either to have neglected the possibility that the parent `constraint` element might bind the namespace in question to two different prefixes, or decided that in that case I want the transform to fail.  That may not be the best strategy for public code written by people with a different sense of what makes sense.  To protect against that case, I'd now change the `xsl:value-of/@select` value to `concat($prefix[1], ':')`.

At this point, the existing code gives up; my local fix tries a couple of additional fallback fixes, first by taking any prefix actually bound to the namespace in the current document:

          
          <xsl:when test="namespace::*[. = $myns][. != '']">
            <xsl:variable name="prefixes" as="xs:string+"
                          select="namespace::*[. = $myns]
                                  [. != ''][1]"/>
          </xsl:when>

If that also fails, my local fix falls back to the technique used in the current code (and in PS's fix) and takes the first prefix specified for this namespace in any enclosing `tei:schemaSpec` element.

             
          <xsl:when test="ancestor::tei:schemaSpec//sch:ns[@uri=$myns]">
            <xsl:value-of
                select="concat((ancestor::tei:schemaSpec//sch:ns[@uri=$myns])[1]/@prefix,':')"/>
          </xsl:when>

If that also fails, we lose and kill the transform (at this point, except for formatting and the nested `choose`, we are back with the existing code).

          <xsl:otherwise>
            <xsl:message terminate="yes"
                         >schematron rule cannot work out prefix for
            <xsl:value-of
                select="ancestor::tei:elementSpec/@ident"/></xsl:message>
          </xsl:otherwise>
        </xsl:choose>
      </xsl:for-each>
    </xsl:function>

sydb:
Whoo-hoo! Will try to take a careful look at @cmsmcq’s suggestion and @peterstadler’s original PR (unless he says not to bother) before Monday.

sydb:
OK. First, some observations that may or may not prove interesting or problemiatc:
* This function is defined 3 times in the Stylesheets/ tree in the dev branch. Except for whitespace, they are identical. (Including that they all have the same signature, taking one argument that should be a node, and returning an xs:string; thank goodness.) Correspondingly, in this PR @peterstadler has proposed his fix once for each of these files:
  * common/functions.xsl
  * odds/extract-sch.xsl
  * odds/odd2odd.xsl
* The program odds/extract-sch.xsl is obsolete, and probably should be deleted. (See TEI issue [#1863](https://github.com/TEIC/TEI/issues/1863), about which @peterstadler is correct, it should have been in the Roma repository, so it may have been moved by the time you read this.)
* This function does not exist in odds/odd2odd.xsl in my radd branch. That’s because it is not called from anyplace in the odds/odd2odd.xsl file. So we are probably OK just ignoring that one. (If I have my way, it won’t exist for much longer, anyway. :-)
* Not surprisingly, I completely concur with both @peterstadler and @cmsmcq’s analysis of the problem.
* It is interesting to note why (I think) I have never run into this particular bug myself. My commandline Schematron procesor of choice (probatron) dies miserably (i.e., aborts with an error message that is entirely unhelpful) if there are two `<sch:ns>` declarations with the same value of `@prefix`, even if they also have the same value of `@uri`. So I routinely do the opposite of what @cmsmcq does, I put in effort to make sure there are no duplicates.
* So, if we ignore odds/extract-sch.xsl and odds/odd2odd.xsl, the only file form which this function is called is odds/teiodds.xsl.
* The only template in that file that calls this function is named "processSchematron", and it is quite a problematic template, IMHO. (E.g., it is called for non-Schematron children of `<constraint>`, has no comments, and has an entire `<xsl:choose>` that could more easily be expressed as an `<xsl:if>`; and does an empty `<xsl:text/>` do anything?) But for now suffice it to say it calls this function twice, once from code that handles Schematron 1.5 (and thus I think we can ignore, as it should be removed) and once that handles ISO Schematron.
* It is being used in the context of an `tei:elementSpec//sch:report` or `tei:elementSpec//sch:assert` to generate the output `sch:rule/@context`.
* Since this is being called when processing Schematron stuff in an ODD (not necessarily a customization ODD), I agree with @cmsmcq that it makes sense to look at the namespace nodes in force. (And it may be obvious that I feel that way, as I used them in extract-isosch.xslt. Although I bet @cmsmcq’s approach is better, and even better once tweaked, below.)

So overall, although @peterstadler’s PR will fix the immediate problem, I think we should implement @cmsmcq’s more general soltion.

However, except for the fact that it is a function and (correctly, IMHO) uses `@as`, @cmsmcq has written in XSLT 1, not XSLT 2. Since an XSLT 2 processor is not required to support the namespace:: axis, I am going to tweak this to be more XSLT 2 -like, using the functions that replace namespace:: and value comparison operators instead of general comparison operators. A few other repairs along the way, too.

I have just checked in changes to the issue366 branch. I have tested that they build P5 locally w/o an error message, but have not tested that they actually build P5 correctly. :-)

Might be a good idea to look at it a little more closely before issuing PR. (I don’t know how confident one is supposed to be of one’s fix before issueing PR. :-)

peterstadler:
Many thanks @sydb and @cmsmcq for fleshing out my simple hack :) 

The tests are currently failing due to ff9837e7d4e22a3606f6db3a3ff14f12cf8609d4 so we should sort that out before merging I'd say.

cmsmcq:
If you want to eliminate all uses of general comparators where value comparators could be used (some people seem to think this is a good idea), you probably want to change line 1314 from `$myns='http://www.tei-c.org/ns/1.0'"` to ... `eq` ....  If `$myns` is not singleton, then the input has nested `elementSpec` elements and is invalid.

-

peterstadler:
now that the current dev branch is building nicely, may I rebase this PR or is someone still working on it?

peterstadler:
seems good to me --  @sydb shall I do the merge?

sydb:
@peterstadler — feel free; I will if you haven’t when I get back in ~2 hours

PR #374, opened 2019-05-11
Add some text to make this findable using google

stuartyeates:


peterstadler:
makes sense to me -- many thanks for the PR!

PR #377, opened 2019-06-05
Add instructions on how to build documentation

bfirsh:
An issue I ran into when getting started. The hosted version seems to be down.


peterstadler:
Yikes, there are at least two issues:

1. the hosted version is down. A workaround is to use the version from our Jenkins server: https://jenkins.tei-c.org/job/Stylesheets-dev/lastSuccessfulBuild/artifact/dist/doc/tei-xsl/index.html
2. the `make doc` is dependent on the proprietary `oxygen.jar`. Without, it  will not create the full documentation but only process a HTML page from the more or less static TEI input at `Documentation/teixsl.xml`

peterstadler:
That said, I think your PR is an uncontroversial improvement, yet forces us to act on further issues (i.e. update `Documentation/teixsl.xml`) -- many thanks ;)

Issue #312, opened 2018-02-21
Some incorrect relative paths to CSSs in the EPub transformation


AlexJitianu:
Hi,

I'm currently updating the TEI version bundled inside Oxygen XML Editor. While running our tests we discovered some bad paths inside xml\tei\stylesheet\epub\tei-to-epub.xsl
`  <xsl:param name="cssFile">../css/tei.css</xsl:param>
  <xsl:param name="cssODDFile">../css/odd.css</xsl:param>
  <xsl:param name="cssPrintFile">../css/epub-print.css</xsl:param>`
These CSSs are not inside a "css" folder but exists directly in the "stylesheet" folder. When transforming, you get errors like:
`Fatal error during transformation using D:\QA\oxygen\frameworks\tei\xml\tei\stylesheet\profiles\default\epub\to.xsl: Failed to read input file; SystemID: file:/D:/QA/oxygen/frameworks/tei/xml/tei/stylesheet/epub/tei-to-epub.xsl; Line#: 171; Column#: 23`

Please let me know if there are other details that I can offer.

Best regards,
Alex

peterstadler:
Hmm, I 'fixed' some path issues recently at d25fe11d497ceac19bdbb87476cc747c13f8fcb2 which seems to cause these troubles. So, how is this supposed to work?
There is a `css` folder within the Stylesheets' root directory. This one I tried to address with the aforementioned changes to the files `epub/tei-to-epub.xsl` and `epub3/tei-to-epub3.xsl`. (Which works for me, btw. Both when calling `profiles/default/epub/to.xsl` and `epub/tei-to-epub.xsl`)

I'm puzzled right now …

AlexJitianu:
Hi,

I think I understand what's happening. There is indeed a css folder insdie the Stylesheets' root directory but it doesn't get included inside the release:
https://github.com/TEIC/Stylesheets/releases/download/v7.44.0/tei-xsl-7.44.0.zip

peterstadler:
I see …
In https://github.com/TEIC/Stylesheets/blob/c1b8f67fbbd350f1ffaa309c30e968796caacb15/Makefile#L73 the css files (from the css subfolder) get copied to the root of the distributable archive. Hence, we end up with different relative paths for these files when working with a checked out repository or with the released archive.
I wonder what the rationale behind this was/is?

[...]

PR #394, opened 2019-09-20
change to full URLs for CSS imports

peterstadler:
This PR fixes the issue #312 with incorrect relative paths to CSSs in the EPub transformation by simply providing full URLs.
Since those files get included via `unparsed-text()` it is not possible to a) use xml catalogs (see https://www.saxonica.com/html/documentation/sourcedocs/xml-catalogs.html) and b) to follow redirects.

### DIFF
--- a/epub3/tei-to-epub3.xsl
+++ b/epub3/tei-to-epub3.xsl
@@ -68,9 +68,9 @@ of this software, even if advised of the possibility of such damage.
   <xsl:param name="bottomNavigationPanel">false</xsl:param>
   <xsl:param name="coverimage"/>
   <xsl:param name="createanttask">false</xsl:param>
-  <xsl:param name="cssFile">http://www.tei-c.org/release/xml/tei/stylesheet/tei.css</xsl:param>
-  <xsl:param name="cssODDFile">http://www.tei-c.org/release/xml/tei/stylesheet/odd.css</xsl:param>
-  <xsl:param name="cssPrintFile">http://www.tei-c.org/release/xml/tei/stylesheet/epub-print.css</xsl:param>
+  <xsl:param name="cssFile">https://www.tei-c.org/release/xml/tei/stylesheet/tei.css</xsl:param>
+  <xsl:param name="cssODDFile">https://www.tei-c.org/release/xml/tei/stylesheet/odd.css</xsl:param>
+  <xsl:param name="cssPrintFile">https://www.tei-c.org/release/xml/tei/stylesheet/epub-print.css</xsl:param>
   <xsl:param name="debug">false</xsl:param>
   <xsl:param name="directory">.</xsl:param>
   <xsl:param name="doctypePublic"/>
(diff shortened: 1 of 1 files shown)


peterstadler:
pinging @raffazizzi and @sydb …

raffazizzi:
Can't imagine any issue with this unless we move away from https, which is unlikely.
Re: your reference to the destination of the CSS files in the release being a can of worms, I think what you suggest is the best solution for the time being.

raffazizzi:
The other issue I can think of is that we're requiring internet access for the TEI bundle to work properly, but again, this may not be avoidable without a more involved rethinking of the release package structure

sydb:
This seems like the worst of the 3 possible solutions:

1. Fix the Makefile so that a css/ subdirectory is created, and the CSS files get put in it
2. Fix the relative paths so they point to the main Stylesheets/ dir, not Styleseets/css/
3. Use absolute URLs.

I realize that if we do (1) we need to see what else breaks, unless we put the CSS files in both places (a bit silly). But that does seem like the best, no?

peterstadler:
@sydb I'm not sure whether you're aware of the original issue (linked above as #312)? 
The problem is that the CSS files are found in different directories in the released (zipped) archive and in the Stylesheets repo. 

For your options 1 and 2 we'd need to reorganise either the directory structure in the repo (and alter the Makefile and ANT scripts and teitoepub3 accordingly) or reorganise the released archive (which would break various links because the directory structure of the archive is unzipped and deployed to the Vault)

sydb:
Yes, @peterstadler, aware of original issue, although I perhaps do not understand it well enough. But besides the obvious questions of why we have a build process that puts the CSS files somewhere other than where they started in the repo, isn’t the problem that oXygen is using [a copy of?] the Stylesheets repo directly, rather than building a release and using that?

martindholmes:
@sydb The oXygen plugin uses a completed build of the TEI products (Stylesheets and TEI), not a copy of the repo.

peterstadler:
Admittedly this PR is a workaround but except its lack of elegance it has no downsides (famous last words). I already tried some approaches (catalog files and redirects) and couldn't think of any other that preserve the current directory structure both in the repo and in the release.
We can turn this into a bigger (new) issue but in the meantime why not merge this PR, please 😍

raffazizzi:
I say we merge this in, but open a ticket about rethinking the release package structure (if there isn't one already) and mention this PR. The goal should be to eventually return to using local paths instead of remote ones.

peterstadler:
The "goal" seems only a minor enhancement to me but yet another ticket floating around with no one caring … (just for the record)

raffazizzi:
@peterstadler you're right about another ticket floating around not being much help. On the other hand I still think forcing users to be connected to the internet to obtain package assets is a step back (unless that already happens elsewhere in the package), so remedying that isn't a minor enhancement. 

Either way, if @sydb doesn't have any other objections by today I'll go ahead and merge.

sydb:
@raffazizzi — today would be jumping the gun a bit from my point of view. (I have meetings all day today, but tend to do a lot of TEI work on Fridays, so hope to look at this and probably merge it tomorrow.)

sydb:
I agree with @raffazizzi that making people have internet connections for no real reason is bad, and thus we should open a ticket (or leave #312 open) about fixing this correctly. That said, I agree with @peterstadler that there is no reason not to merge this in the meantime.

PR #439, opened 2020-06-04
replace http requests/links with https in XSL scripts

### LINKED ISSUE #400: Some XSL libs still point to http://www.tei-c.org/Vault/P5
Some libs have been changed to use https, but others are still pointing to plain http, which will no longer work when trying to retrieve p5subset.xml. Is there any reason not to change all of these?

peterstadler:
some housekeeping, fixes #400


sydb:
@peterstadler — the Test2/ checks fail (at least for me) due to a diff error you might expect:
```
     [exec] output:       <link href="https://www.tei-c.org/release/xml/tei/stylesheet/tei.css"
     [exec] expect:       <link href="http://www.tei-c.org/release/xml/tei/stylesheet/tei.css"
```
for file Test2/outputFiles/testBinaryObjects1.html (vs. Test2/expected-results/testBinaryObjects1.html).

I can probably start fixing these in about an hour.

peterstadler:
@syd are you sure you are on the latest commit? For me, the tests are not failing and Travis also thinks it's good.

sydb:
I do not know if they have an official status at all, let alone what it might be, but the set of Test2/ checks is overall superior to those in Test/, or at least @martindholmes and I think so. And it would be nice to have them working.

peterstadler:
I understand—and I'd be glad if you fixed the failing tests :)

sydb:
Working on fixing it now …

sydb:
Fixed in #49bcba5, I hope.

martindholmes:
@peterstadler Test2 tests often catch things the regular tests don't, and they also tend to fail less frequently on minor pointless stuff because they're more robust against irrelevant changes such as changes to generated ids. We also think they're faster, easier to maintain, better documented, and they actually test what they're supposed to test. Files have meaningful names, and are organized in a logical way, so it's easier to figure out what's going on. Whether Council wants to accept them or not is up to Council, but @sydb and I use them exclusively when we're actually working on ODD stuff.

Issue #440, opened 2020-06-04
docx2tei ignores @w:val attribute on bold and italic text


rvdb:
The docx to TEI conversion is a bit coarse w.r.t. the treatment of italic and bold text. When processing the `<w:i>` and `<w:b>` elements, no checks are made for a `@w:val` attribute. This attribute provides a binary value for switching the style on and off. By consequence, text marked in a Word source file as:

    <w:r>
      <w:rPr>
        <w:i w:val="1"/>
        <w:b w:val="0"/>
        <w:iCs w:val="0"/>
      </w:rPr>
      <w:t>italic, not bold</w:t>
    </w:r>

...should be transformed to:

    <hi rend="italic normalweight">italic, not bold</hi>

Yet, the code in [textruns.xsl](https://github.com/TEIC/Stylesheets/blob/364b9bb592871cd440f78a8aaecb895b2737c8bd/docx/from/textruns.xsl#L250-L272) merely checks for the presence of the `<w:i>` and `<w:b>` elements, thus ignoring possible `w:val="0"` values. The example above hence is wrongly transformed to:

    <hi rend="italic bold">italic, not bold</hi>

Attached is a zip file [italic-bold-test.zip](https://github.com/TEIC/Stylesheets/files/4730536/italic-bold-test.zip) containing three files:

- `italic-bold-test.docx`: a Word file illustrating the Word encoding
- `italic-bold-test-incorrect.xml`: a TEI file exported with Oxgarage, illustrating the incorrect output
- `italic-bold-expected.xml`: a TEI file with expected output

I have a fix, which I'll propose in a PR.

rvdb:
The `@w:val` switch is used on more basic styles: `<w:i>`, `<w:b>`, `<w:strike>`, `<w:dstrike>`, `<w:smallCaps>`, and `<w:caps>`.

Also, `@w:val` seems to allow for a range of binary values: (`0`, `false`, `off`) vs (`1`, `true`, `on`). 

I've expanded the fix for the `basicStyles` template to cover these cases, too. Attached is an updated test file, with incorrect and expected outputs.
[basicStyles-test.zip](https://github.com/TEIC/Stylesheets/files/4744641/basicStyles-test.zip)

PR #441, opened 2020-06-04
fix for handling @w:val[. = "0"] on //w:i and //w:b elements

rvdb:
fixes https://github.com/TEIC/Stylesheets/issues/440


peterstadler:
@rvdb I just rewrote the history of this branch by going back to your latest commit 215cd4d9fb6cfe7a50864587917980961f4e407b and then rebasing the branch onto current dev. Please do not try to update this branch again (or do any other work within this (=your) branch) since this will probably bring more trouble due to the rewritten history. 

When the tests pass I will merge the PR and we are set :-)

rvdb:
@peterstadler I'll delete it right away. Sorry for the mess, and thanks for cleaning it up. Is this how I should have done it in the first place: rebase instead of pull?

peterstadler:
No, it's hard to tell now what exactly went awry. Anyway, I was just looking back at the issue #440 and found that you were also providing tests which I'd like to add to this PR. Am I right that I only need the second zip file with the basicStyles?

rvdb:
@peterstadler Yep, the second version covers tests for all modified styles in this PR.

PR #448, opened 2020-07-03
Subset setup

martindholmes:
Implementation by @sydb and me of our decisions during the meeting 2020-07-03 on build circularity.


Makefile
```diff
@@ -119,11 +119,12 @@ dist: clean release
 release: common doc oxygendoc build profiles
 
 installxsl: build teioo.jar
-	mkdir -p ${PREFIX}/share/xml/tei/stylesheet
+	mkdir -p ${PREFIX}/share/xml/tei/stylesheet ${PREFIX}/bin ${PREFIX}/source
 	(tar cf - lib teioo.jar) | (cd ${PREFIX}/share/xml/tei/stylesheet; tar xf - )
 	(cd release/xsl; tar cf - .) | (cd ${PREFIX}/share; tar xf  -)
-	mkdir -p ${PREFIX}/bin
-	cp bin/transformtei ${PREFIX}/bin
+	cp --preserve=timestamps bin/transformtei ${PREFIX}/bin
```
raffazizzi, line 125: Why preserve the timestamp?

martindholmes: @raffazizzi Good question, and we had a good reason, but I can't remember it. @sydb?

hcayless (approved):
This looks good to me. In the Stylesheets meeting we discussed adding a message to `transformtei` that gives the current version of the p5subset used instead of the warning message that a remote copy will be retrieved. I will open an issue for that.

Issue #436, opened 2020-05-26
JTEI OpenEdition output should enable URL wrapping


martindholmes:
When a URL is both the text of a link and the target (`@href`), which is common in e.g. footnotes which consist only of URLs, browsers have problems with justification because they see the URL as a single word. One solution to this is to detect when the text of a link contains a URL, and insert zero-width breaking spaces at URL component boundaries (slashes and periods), so the browser has lots of options for wrapping it.

martindholmes:
I believe the pull request fixed this, so I'm closing it.

PR #450, opened 2020-07-08
added zero-width spaces after slashes and periods in URLs that are be…

rvdb:
…ing output in the text

Addresses https://github.com/TEIC/Stylesheets/issues/436


profiles/jtei/openedition/to.xsl
```diff
@@ -674,7 +674,20 @@
...
+               Note: in order to avoid breaking in between repeated slashes and periods, and within 
+               '../',  this replacement has to happen in two passes.
+          -->
+          <xsl:value-of select="
+            replace(
+              replace(
+                @target,
+                '([/])([^/])',
+                '$1&#8203;$2')
+            , '([.])([^./])',
```
martindholmes, line 688: @rvdb Not strictly necessary, but would it be a good idea to have a backslash before the periods in `([.])([^./])` to make it clear they're literal periods? I think it works both ways in XPath, but it might help readability.

rvdb: @martindholmes Sure, I always tend to forget that it works both ways in XPath, and that the version without backslash is an XPath syntax quirk. I've added them, thanks for reviewing.

martindholmes: Thanks Ron. I've given myself so many hours of misery forgetting those backslashes in less forgiving languages. :-)

sydb: I have to say I disagree with @martindholmes completely. I think `[.]` and `[^./]` are simply shorter and easier to read than `[\.]` and `[^\./]` (which looks like it means everything except backslash, dot, or slash). That said, I don’t understand why we would use `[.]` instead of just `\.`. But more importantly, I think the stakes here are so small I would not be able to find them with a magnifying glass.
But what less forgiving languages do you have in mind? Certainly Perl, grep, egrep, and Emacs are all happy with just `[.]`, no?

martindholmes: I think in this context yes, but there are certainly cases in JavaScript where I've caused myself pain by forgetting the backslash. So my policy is: if i _mean_ a literal period, and a backslash is _allowed_, I use the backslash.

Issue #454, opened 2020-08-11
classRef/@except processing fails


martindholmes:
The Stylesheets Working Group, trying to integrate the latest P5 changes into the Stylesheets repo, has discovered that classRef/@except is correctly processed for RNG/RNC, but for DTDs it creates an invalid DTD.

This is triggered by the change here:

https://github.com/TEIC/TEI/commit/193e6283321aeeb68b9a93b0d95b10a82148ef74

which closes ticket https://github.com/TEIC/TEI/issues/1848.

lb42:
This looks the same as a problem reported in jan 2017 (#220)
 That ticket also includes a useful testfile.

sydb:
Thanks, @lb42. Helpful test file. However, I do not think this is the same problem, though clearly related. 

This problem seems to be rooted, I have discovered (after several hours), in the CLASSMEMBERS key, defined in odds/teiodds. That key only matches `<classSpec>` elements, not `<classRef>`. I have fixed it (and CLASSMEMBERS-CLASSES, which has the same problem), and that has fixed the immediate problem. However, testing what else might go wrong will take quite a while. Once tested somewhat I will push up a new branch, **sydb-454** or some such.

hcayless:
When this is fixed, the content models of `<lem>` and `<rdg>` will need to be updated.

ebeshero:
Is this related to #754 ?

sydb:
> Is this related to [#754](https://github.com/TEIC/Stylesheets/issues/754) ?

I do not think so, but not at all sure. It is, however, probably related to the failure of the obvious solution to TEIC/TEI#2460 — just adding `except="address"` to either `<classRef key="model.phrase">` or `<classRef key="model.divBottom">` seems to cause the entire class reference to be ignored.

Issue #220, opened 2017-01-07
ODD to DTD problems #2


lb42:
A DTD must contain 5 parameter entities for each model class defined, with names consisting of the model class name suffixed by each of the  "methods" sequence, alternate etc. The declarations look like this
~~~~
<!ENTITY % testClass "x |y |z"> 
<!ENTITY % testClass_sequence "x, y, z"> 
<!ENTITY % testClass_sequenceOptional "x?, y?, z?"> 
<!ENTITY % testClass_sequenceOptionalRepeatable "x*, y*, z*"> 
<!ENTITY % testClass_sequenceRepeatable "x+, y+, z+"> 
~~~~
However, an ODD declaration like this 
~~~~
      <elementSpec ident="testP3">
          <content>
            <classRef key="testClass" expand="alternate" except="z"/>
          </content>
        </elementSpec>
~~~~
generates DTD like this
~~~~
      <!ELEMENT testP3 (((%testClass_alternate;)))>
~~~~
which is wrong because there is no such parameter entity, independently of being inadequate because it ignores the @except attribute. 

See further the attached test file
[testClass.zip](https://github.com/TEIC/Stylesheets/files/691693/testClass.zip)

martindholmes:
What results do you get if you convert the ODD to RelaxNG and then run Trang to get a DTD?

lb42:
Trang gives me a shed load of apologetic error messages saying "sorry cannot handle this kind of oneormore" and no output. Which I suppose is fair enough.

lb42:
Interestingly, there was some code in oddtodtd.xsl which tried to handle these classRefs properly, but it was simplified out of existence at  04893050b67b8f87112fb7ee4dca0a003397453e @sebastianrahtz sebastianrahtz committed on 19 Feb 2014

PR #455, opened 2020-08-13
update Test/Makefile so it is a bit more understandable

sydb:
I _think_ this is ready to rock and roll. It has the new p5subset from TEI dev branch in it. All tests (in Test/) pass, as far as I can tell. (I have not tried Test2/ yet; maybe can do that after dinner.)

If one or two of you, the requested reviewers, can check out this branch and test the test procedure, that would be appreciated. If you `cd Stylesheets/Test` and then `make help` you will get some brief documentation.


sydb:
P.S. This means I do not know what went wrong with the build & test we tried at the Tue 11 Aug meeting, but that particular test is passing now.

peterstadler:
Thank you @sydb for all the work but it's really hard to review since you altered the build process, the source for the tests and the expected results at once. We now had a lot of reverse engineering to do if we wanted to really check this!

peterstadler:
I tried to run the tests with your modified Makefile against our current `p5subset.xml` and expected results in the dev branch. 

```
		then diff  -r  actual-results/testnotes/ expected-results/testnotes; fi
diff -r actual-results/testnotes/index.html expected-results/testnotes/index.html
13c13
<     <link href="../../../tei.css" rel="stylesheet" type="text/css" />
---
>     <link href="../../tei.css" rel="stylesheet" type="text/css" />
diff -r actual-results/testnotes/one.html expected-results/testnotes/one.html
13c13
<     <link href="../../../tei.css" rel="stylesheet" type="text/css" />
---
>     <link href="../../tei.css" rel="stylesheet" type="text/css" />
diff -r actual-results/testnotes/three.html expected-results/testnotes/three.html
13c13
<     <link href="../../../tei.css" rel="stylesheet" type="text/css" />
---
>     <link href="../../tei.css" rel="stylesheet" type="text/css" />
diff -r actual-results/testnotes/two.html expected-results/testnotes/two.html
13c13
<     <link href="../../../tei.css" rel="stylesheet" type="text/css" />
---
>     <link href="../../tei.css" rel="stylesheet" type="text/css" />
make[1]: *** [test-to-html] Error 1
make: *** [test] Error 2
```

Seems some paths are wrong now?

peterstadler:
ok, did more testing: the paths are not 'wrong' but reflect the new folder structure that @sydb applied, i.e. the extra path element 'actual-results'.
Running the modified Makefile against the current Stylesheets-dev version of `p5subset.xml` and expected results did not produce any other serious differences so I confirm that @sydb did an excellent job in refactoring the Test-Makefile.

Yet, when I replace the source/p5subset.xml with the current one from https://jenkins.tei-c.org/job/TEIP5-dev/lastSuccessfulBuild/artifact/P5/release/xml/tei/odd/p5subset.xml, I still get the DTD error 
```
<!ELEMENT lem (#PCDATA|_DUMMY_model.divLike|%model.divPart;|titlePage|||_DUMMY_m
                                                                      ^
```

source/p5subset.xml
```diff
@@ -46036,6 +46002,9 @@ reading groups, each associated with witness specifications.</p>
         <textNode/>
         <classRef key="model.divLike"/>
         <classRef key="model.divPart"/>
+        <elementRef key="titlePage"/>
+        <classRef key="model.titlepagePart"/>
```
peterstadler: @sydb you're cheating and removed the `except="binaryObject graphic"`

peterstadler (changes requested):
My final conclusion: The current PR is a refactoring of the Makefile with adjusted tests but it does not solve the problem reported at #454 and #220

source/p5subset.xml
hcayless: So, to be fair, it's possible to solve this problem by just pointing directly at the desired elements in `<lem>`'s content model. I was naively assuming that the Stylesheets implemented that feature of Pure ODD. They don't, really.

sydb:
Sorry I was not available to reply earlier to your comments starting 8 hours ago, @peterstadler; might have been able to help you avoid some of that journey of discovery you went through. Such are time zones. Anyway, thanks for verifying.

source/p5subset.xml
sydb: Oh BLEEP!
I must have copied the p5subset we were playing with on Tue morning, not the p5subset from Jenkins or my local dev build. Thank you for catching that, @peterstadler! (No wonder things worked so well. :-)
Yes, @hcayless, our conclusion Tue morning was that this is probably an error in handling a `<classRef>` with `@except`, and thus it could either be fixed (best plan) or worked around by tweaking the ODD (probably what we will have to do for intermediate hack for this release until we can get a real fix).
I hope to look at the `@except` problem starting in perhaps 10 mins.

sydb:
It seems I am insufficiently savvy with GitHub to figure out how to see the actual change you requested, @peterstadler, but I presume it is to swap in the current p5subset from dev. (Which has a lot more changes than just the missing `@except`.)

I will try to do that in the next hour.

sydb:
@martindholmes has convinced me that, despite @peterstadler’s heroic efforts, the fact that I (inadvertently) combined an update to p5subset with the changes to the Makefile make this PR too hard to properly review in a timely manner.
Thus I am suggesting we defer full review and pulling of this branch until after release. (I personally plan to use this new Makefile as I work on #455, but that only means there will be different testing on it from me and Jenkins, which is good.)

sydb:
OK, @martindholmes, and any other requested reviewer (@hcayless, @raffazizzi, @jamescummings, @lb42), I think it would be good for at least 1 other person to verify this works.

Note that what you are testing here is **Tests/Makefile**. You should feel free to swap in whatever p5subset.xml you want (by changing **source/p5subset.xml**, or by setting a default source on the commandline, for exampe `make DEFAULTSOURCE=/home/duck/quack/p5subset.xml`), and to make whatever changes to expected_results/ you want in order to see how the Makefile works. Since the directory structure of the files generated by the Makefile are different (due to new actual_results/ directory) it is a given that some files in expected_results/ will be different, because the path to (e.g.) CSS files is different.

When we actually merge this PR in, there is no reason to merge those files (subset and expected_results/), only the Makefile itself. A separate commit could deal with expected_results/.

It is worth testing both the standard fix-one-file-at-a-time procedure that some of us get frustrated with, and the suggested all-at-once procedure. To see the latter issue `make help`.

martindholmes:
@sydb The tests work fine, but the p5subset in this branch is way different from the current one in dev, so everything will break when it's merged. I think what should happen is:

1. Pull the current p5subset from the dev branch of P5 into Stylesheets/dev

2. Pull those changes into your branch.

3. Update expected-results etc. as necessary.

4. Merge into dev.

I do like the much more helpful output from a lot of the tests, though!

PR #462, opened 2020-10-06
Make teitoepub3 self-recognize just like teitoepub

craigberry:
teitoepub3 has not been offering all the same command-line options
as teitoepub because the checks for whether those options were
relevant were doing an equality comparison with "teitoepub" and
obviously "teitoepub3" fails that comparison. So make the checks
match either one (with or without the "3" at the end).

### DIFF
--- a/bin/transformtei
+++ b/bin/transformtei
@@ -48,17 +48,19 @@ case $format in
 	;;
 esac
 
-if [ $script = "teitoepub" ]
-then
+case "$script" in
+    teitoepub*) # could be teitoepub or teitoepub3
 echo "  --fileperpage        # create output files one per page "
 echo "  --nocompress	     # disable compression of epub output file"
-fi
+    ;;
+esac
+
 echo 
 echo "  Options, shown with defaults:"
 echo "  --saxonjar=$SAXONJAR  # location of Saxon jar file"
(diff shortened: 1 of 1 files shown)


sydb:
We (TEI Stylesheets group) have looked at this, and it seems fine. We do want an actual test, just in case, before merging. Thus @ebeshero, @HelenaSabel, and I are going to test it. If all goes well we will just merge. If not we will add “needs discussion” flag.

craigberry:
Thanks very much.  The test that I did was just using the --css option to specify a custom stylesheet with teitoepub3.  You don't get that option before this patch.  I then generated about 52,000 ePubs from EEBO-TCP-derived texts (actually twice, once in original spelling and once in standard spelling) and used Calibre to generate PDFs and MOBIs from the ePub files using the code [here](https://bitbucket.org/eplib/downloads/src/master/). This is not to dissuade you from further testing, but just to provide context about what's already been done.

HelenaSabel:
Thank you for the info @craigberry!

I tested both `teitoepub` and `teitoepub3` and it worked well. I also tested without problem different combinations of the following parameters:
 - `--css`
 - `--coverimage`
 
The epub readers I used to check the results were Calibre and Sigil.

Issue #463, opened 2020-10-07
Processing of `@rendition` attributes in HTML is namespace-ignorant


martindholmes:
This pair of xsl:keys in html/html_header.xsl:

~~~~~
<xsl:key name="ALL-EXTRENDITION" match="@rendition[not(starts-with(.,'simple:') or starts-with(.,'#'))]" use="1"/>
  <xsl:key name="EXTRENDITION"
	   match="@rendition[not(starts-with(.,'simple:') or starts-with(.,'#'))]" use="."/>
~~~~~

matches `@rendition` attributes in any namespace. That means it matches `@rendition` attributes in the `<egXML>` examples namespace, when it shouldn't. The result is that if you try to process an ODD file into HTML documentation, and your ODD file happens to contain examples which include the use of a private URI scheme in `@rendition` attributes, the conversion will fail with an "unknown protocol" error. This simple fix solves the problem:

~~~~~
  <xsl:key name="ALL-EXTRENDITION" match="tei:*/@rendition[not(starts-with(.,'simple:') or starts-with(.,'#'))]" use="1"/>
  <xsl:key name="EXTRENDITION"
	   match="tei:*/@rendition[not(starts-with(.,'simple:') or starts-with(.,'#'))]" use="."/>
~~~~~

I'll create a pull request for the fix and link it below.

ebeshero:
@martindholmes ~~This is strange.~~ I've been testing with the bleeding edge plugin in oXygen, and first just trying to see if I can reproduce the error. ~~I haven't been able to yet: I'm getting good HTML documentation and not seeing errors, not even any any validation errors when I inspect the HTML.~~ **Correction:** Yes: the error is generated, but oXygen still generates the HTML documentation and that documentation looks okay.

 I tried two different `<elementSpec>`s with some foreign namespace content in egXMLs in my simple ODD: 

One elementSpec was for an element in the TEI namespace:
```
 <elementSpec ident="seg" module="linking" mode="change">
               <attList>
                  <attDef ident="rendition" mode="replace" usage="req">
   
                     <valList type="closed">
                        <valItem ident="rnd:upover"/>
                        <valItem ident="rnd:downlow"/>
                     </valList>
                  </attDef>
               </attList>
               <exemplum xml:lang="en">
                  <p>The <gi>seg</gi> is here to help me test stuff that isn't working in ODD transformations. </p>
[...]

PR #464, opened 2020-10-07
Add namespace sensitivity to xsl:keys used to process @rendition per …

martindholmes:
Pull request for fix for issue #463.

### DIFF
--- a/html/html_header.xsl
+++ b/html/html_header.xsl
@@ -53,19 +53,26 @@ of this software, even if advised of the possibility of such damage.
       </desc>
    </doc>
 
-  <xsl:key name="ALL-EXTRENDITION" match="tei:*/@rendition[not(starts-with(.,'simple:') or starts-with(.,'#'))]" use="1"/>
+  <xsl:key name="ALL-EXTRENDITION"
+	   match="@rendition
+		  [not( ancestor::teix:* )]
+		  [not( starts-with( ., 'simple:') or starts-with( ., '#') )]"
+	   use="1" />
   <xsl:key name="EXTRENDITION"
-    match="tei:*/@rendition[not(starts-with(.,'simple:') or starts-with(.,'#'))]" use="."/>
+	   match="@rendition
+		  [not( ancestor::teix:* )]
+		  [not( starts-with( ., 'simple:') or starts-with( ., '#') )]"
+	   use="." />
   <xsl:key name="ALL-LOCALRENDITION" match="tei:rendition" use='1'/>
 
   <doc xmlns="http://www.oxygenxml.com/ns/doc/xsl">
-      <desc>Process element teiHeader</desc>
-   </doc>
+    <desc>Process element teiHeader</desc>
+  </doc>
   <xsl:template match="tei:teiHeader"/>
 
   <doc xmlns="http://www.oxygenxml.com/ns/doc/xsl">
-      <desc>make a style section from rendition elements in the header</desc>
-   </doc>
+    <desc>make a style section from rendition elements in the header</desc>
+  </doc>
   
   <xsl:template name="generateLocalCSS">
       <xsl:if test="key('ALL-LOCALRENDITION',1)">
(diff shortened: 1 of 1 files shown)


ebeshero:
That fix is admirably simple! The way to test this I think will be to make an ODD with some customization in a distinct namespace, and see how the HTML turns out. I'll try it out here...

martindholmes:
@ebeshero Sorry, I should have provided a test file to demonstrate the problem. It just has to be an ODD file with some documentation like this in it:

~~~~~
<egXML xmlns="http://www.tei-c.org/ns/Examples">
  <seg rendition="rnd:upover">g</seg>
</egXML>
~~~~~

The "rnd" prefix will cause an undeclared prefix error, which will go away if the fix is applied.

ebeshero:
@martindholmes I've confirmed your fix works on my testerODD just by running `odd2html.xsl` on dev and then on your branch, so I'm satisfied! But I think @sydb suggests a slightly modified solution in #463.

ebeshero (approved):
@martindholmes I've confirmed your fix works on my testerODD just by running `odd2html.xsl` on dev and then on your branch, so I'm satisfied! But I think @sydb suggests a slightly modified solution in #463.

martindholmes:
@ebeshero Yes, I think @sydb's version is better, so I'll let him update the pull request (if he wants to) or tell me to update it accordingly.

sydb:
I don’t think I know how to update a pull request. Do I just commit a change to the issue_463_mdh_rendition_ns branch?

martindholmes:
@sydb I think if you commit to that branch, then push, the pull request should be updated automatically.

sydb:
OK, will try, hopefully this afternoon.

sydb:
Pushed 95ffcb2, which implements the fix I mentioned on [the ticket](#463). Let’s see what happens to this PR.

However, my fix uses `ancestor::teix:*`. Is there any reason to look up the ancestor tree? I mean, in the completely bonkers case of
~~~~~xml
<egXML xmlns="http://www.tei-c.org/ns/Examples">
  <p rendition="#duck">Quack, quack!</p>
  <dbk:para rendition="#duck">Waddle, waddle!</dbk:para>
</egXML>
~~~~~
should the `@rendition` on the DocBook paragraph be in the key or not? If yes, or if it doesn’t matter, then we should use parent:: instead of ancestor::, no?

martindholmes:
@sydb ancestor includes parent, no? I think it is important to account for all the instances where @rendition might appear on an element which is not in the tei namespace.

martindholmes:
@sydb 's commit is now part of the pull request.

sydb:
Sorry, @martindholmes, I did not see your 1st comment yesterday, to wit:

> ancestor includes parent, no? I think it is important to account for all the instances where `@rendition` might appear on an element which is not in the TEI namespace.

Yes, the 1 element on the `parent::` axis (we know it is 1, not 0, because the current node is an attribute, namely `@rendition`) is a subset of all (the 4 or more) elements on the `ancestor::` axis.

But I don’t know what “accounting for an instance” means: does it mean include in the EXTRENDITION keys or exclude from them?

Of course, one thing to find out is whether `@rendition="prefixed:value"` throws an error when it is on a DocBook element in a `<teix:egXML>` (or on the `<teix:egXML>` itself).

I hope to test that this morning, but it depends a bit on how much time I end up fighting with CSS for the WWP presentations. :-)

martindholmes:
@sydb The intention is to exclude any `@rendition` attributes which are not on elements in the TEI namespace from the xsl:keys because the process should not attempt to do anything with them at all; by definition it knows nothing about them. To be more specific, I believe the keys should only include `@rendition`s in the empty namespace whose parent element is in the TEI namespace. Those are the `@rendition` attributes on which TEI processing should operate; all others (including any in egXMLs, which were the immediate cause of finding the bug) are not the business of the processor at all.

sydb (approved):

PR #466, opened 2020-10-13
align list processing in docx2tei with GL prose: differentiate betwee…

### LINKED ISSUE #465: docx2tei conversion needs to differentiate between list/@type and list/@rend
I've noticed how the docx2tei conversion still outputs renditional features of lists as the `list/@type` atribute, e.g.:

    <list type="ordered">
      <item>...</item>
    </list>

Hence, this processing is out-of-sync with the Guidelines, which since release 2.7.0 differentiate between `@type` and `@rend` (see https://sourceforge.net/p/tei/bugs/460).  The correct output for an ordered list should be:

    <list rend="ordered">
      <item>...</item>
    </list>

I'll propose a PR.

rvdb:
…n list/@type and list/@rend

(as implemented in TEI-2.7.0 of the GL, see https://sourceforge.net/p/tei/bugs/460)

addresses https://github.com/TEIC/Stylesheets/issues/465

### DIFF
--- a/Test/expected-results/test-rtf2tei.xml
+++ b/Test/expected-results/test-rtf2tei.xml
@@ -58,7 +58,7 @@
       <p>Na <ref target="http://nl.ijs.si/e-zrc/rtf2tei/">http://nl.ijs.si/e-zrc/rtf2tei/</ref> najdete servis za pretvorbo Word datotek v <ref target="http://www.tei-c.org/Guidelines/P5/">TEI P5</ref>. Spletna stran vsebuje tudi navodila za uporabo same spletnega servisa, pričujoče besedilo pa vsebuje navodila za oblikovanje Word datotek za servis in obenem služi kot testno besedilo, ki ga lahko pošljete servisu, da vam vrne ustrezno stavljen rezultat v XML oz. HTML. </p>
       <div>
         <head><anchor xml:id="Toc317600600"/>Splošna navodila</head>
-        <list type="ordered">
+        <list rend="numbered">
           <item>Besedilo naj uporablja naslove (Heading 1, Heading 2, itn.), saj bo le tako pravilno razčlenjeno na razdelke (&lt;div&gt;) v TEI. Besedilo pred prvim naslovom gre v element &lt;front&gt;. <lb/><hi rend="underline">Namig</hi>: Kaj točno so naslovi lahko preverite preko »prikaži ¶«. Pravilni naslovi imajo na začetku vrstice napisan kvadratek, ▪ <lb/><hi rend="underline">Pozor</hi>: »Naslovi« lahko pomotoma postanejo tudi prazne vrstice ali pa prelomi strani, kar preverite skozi kvadratek. Ali se naslovi res pravilno pretvorijo, preverite v kazalu generiranega HTML, pa tudi v sami datoteki Word lahko naredite kazalo.</item>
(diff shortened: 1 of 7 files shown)


martindholmes:
@rvdb Thanks for this. We'll have to fix the tests as well, but that shouldn't be difficult. Should the other recommended values of list/@type also be included, as well as "gloss"?

~~~~~
gloss
    each list item glosses some term or concept, which is given by a label element preceding the list item.
index
    each list item is an entry in an index such as the alphabetical topical index at the back of a print volume.
instructions
    each list item is a step in a sequence of instructions, as in a recipe.
litany
    each list item is one of a sequence of petitions, supplications or invocations, typically in a religious ritual.
syllogism
~~~~~

rvdb:
@martindholmes They probably should, but I didn't have the time to figure out where the 'dl' list definitions were coming from in the first place. So, I guess this could be solved by modifying either (or both):

- the function `tei:get-listtype()` in `functions.xsl`  
https://github.com/TEIC/Stylesheets/blob/dev/docx/from/functions.xsl#L184-L206
 - the last `<choice>` block in the named template `listType` in `lists.xsl` 
https://github.com/TEIC/Stylesheets/blob/dev/docx/from/lists.xsl#L174-L179

(but I don't know how those other list types should be derived from the input)

sydb:
@rvdb : Can you supply a .docx test file that has lists that get incorrectly changed by the standard Stylesheets?
Thanks.
— @martindholmes & @sydb

martindholmes:
That function tei:get-listtype() should be changed to:

~~~~~
<xsl:function name="tei:get-listtype" as="xs:string">
        <xsl:param name="style"/>
        <xsl:choose>
            <xsl:when test="starts-with($style,'dl')">
                <xsl:text>gloss</xsl:text>
            </xsl:when>
            <xsl:when test="starts-with($style,$ListBullet)">
                <xsl:text>bulleted</xsl:text>
            </xsl:when>
            <xsl:when test="starts-with($style,$ListContinue)">
                <xsl:text>bulleted</xsl:text>
            </xsl:when>
            <xsl:when test="starts-with($style,$ListNumber)">
                <xsl:text>numbered</xsl:text>
            </xsl:when>
            <xsl:when test="$style=$List">
                <xsl:text>numbered</xsl:text>
            </xsl:when>
            <xsl:otherwise>
                <xsl:text/>
            </xsl:otherwise>
        </xsl:choose>
    </xsl:function>
~~~~~

I'm not sure about `$ListContinue`, and I think one of them should probably be "simple", but I'm not sure. We need someone who knows docx.

rvdb:
Hi @sydb ,

Here's a test file illustrating the original issue (https://github.com/TEIC/Stylesheets/issues/465): [list-rend.docx](https://github.com/TEIC/Stylesheets/files/5431479/list-rend.docx)


When that file is being converted to a TEI P5 file with Oxgarage, you'll notice how the `@type` attribute is used, instead of `@rend`, for displaying the "unordered", resp. "ordered" rendition styles. Which is out of sync with the Guidelines prose.

I have no experience with other list types, though, and don't even know where e.g. that "dl" value would come from in docx (custom styles?).

martinascholger:
Comment from Stylesheets group meeting: @peterstadler is updating the expected results and running the tests (including test2).

peterstadler:
local tests run successful now, so I'll go ahead and merge

Issue #424, opened 2020-02-25
`@ns` in ODD customizations causes horrible problems in RNG


ebeshero:
We find that when an `@ns` attribute appears on a schemaSpec or an elementSpec (whether as a result of being generated as a default attribute by oXygen, or for some other reason), it results in broken RNG's that don't carry customizations of attributes in particular. 

What we've tested at the Stylesheets meeting on 2020-02-25: All of these results are coming from the RNC output from trang. (Stylesheets use trang.)

1) attDef with/without the `@ns` set to TEI doesn't do anything, which means it's inconsequential to the processing. (But it should be creating an attribute in the TEI namespace.)
1) attDef with an `@ns` set to a non-TEI namespace, with a schemaSpec set to a different non-TEI namespace results in attDef's `@ns` gets ignored and the schemaSpec's `@ns` prevails.
1) elementSpec and schemaSpec with default `@ns`set to TEI namespace results in suppressing any customized attDef from the RNG.
1) elementSpec and schemaSpec with default `@ns`set to something other than the TEI namespace results in GOOD results: element and attribute customizations come through. (We should check and see if their customized namespaces appear in the RNG.)
1) elementSpec with `@ns` of TEI and schemaSpec without `@ns` is fine. 
1) schemaSpec with `@ns`of TEI and elementSpec without `@ns` works
1) schemaSpec with `@ns`of TEI and elementSpec  `@ns` of TEI suppresses all the attributes customized in attDef
1) elementSpec with `@ns` that's NOT  TEI and schemaSpec with `@ns` of TEI: results in losing the attributes on that element.  
1) attDef with `@ns` other than TEI, and no `@ns` anywhere else works just fine. (Make sure you use `mode="add"` when testing. We might have forgotten to do this in earlier tests.)

Perhaps where the namespaces don't match, the entire customized element is lost  from the RNG?

There are probably about six tickets to be opened from this. 
* There are problems with rng:grammar and Jing.... testing with Trang in Roma Command Line tool. How does this behave in Oxgarage or elsewhere? 
* What's happening in the tei-to-rnc conversion?

Please amend this ticket as we continue testing. People involved today: @sydb @martindholmes @martinascholger @hcayless  and @ebeshero with Nicholas Cole for a moment but we're having trouble finding his GitHub handle!!!!

martindholmes:
On the Trang/RngToRnc issue, I posted this ticket:

https://github.com/TEIC/Stylesheets/issues/423

[...]

PR #468, opened 2020-10-17
Properly handle inherited @ns on elementSpec

dmj:
Partial fix for #424


HelenaSabel:
Drawing inspiration from your proposal, @dmj ,the TEI Stylesheets group has improved the `tei:uniqueName()` function. See [PR 563](https://github.com/TEIC/Stylesheets/pull/563) (work in progress).

PR #475, opened 2020-11-01
Don't use content of tei:index in tei:head

### LINKED ISSUE #382: index, term in head appears in table of contents
Element index is considered out of normal text content; it is allowed in head, where indeed its content does not appear in HTML. But it does appear in the table of contents.
See http://www.d-meeus.be/linux/sandbox/fruits.html transformed from http://www.d-meeus.be/linux/sandbox/fruits.xml with /usr/share/xml/tei/stylesheet/html5/html5.xsl (7.47.0).

dmj:
Fix #382

### DIFF
--- a/common/common_core.xsl
+++ b/common/common_core.xsl
@@ -1171,6 +1171,7 @@ of this software, even if advised of the possibility of such damage.
   <xsl:template match="tei:*" mode="plain">
       <xsl:apply-templates mode="plain"/>
   </xsl:template>
+  <xsl:template match="tei:index" mode="plain"/>
   <xsl:template match="tei:note" mode="plain"/>
   <xsl:template match="tei:app" mode="plain"/>
   <xsl:template match="tei:pb" mode="plain"/>
--- a/common/common_linking.xsl
+++ b/common/common_linking.xsl
@@ -703,7 +703,10 @@ of this software, even if advised of the possibility of such damage.
 		       </xsl:when>
 		       <xsl:when
 			   test="starts-with(local-name(.),'div') and tei:head">
-			 <xsl:value-of select="tei:sanitize(tei:head)"/>
+                         <xsl:variable name="plain-head" as="xs:string*">
+                           <xsl:apply-templates mode="plain" select="tei:head"/>
+                         </xsl:variable>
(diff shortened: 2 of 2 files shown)


sydb:
David — I don’t suppose you have a test case that demonstrates `<index>` inside `<head>` as being problematic?

dmj:
Not yet, I suppose. How does a test case look like? I think I can come up with one for the Makefile based tests, the Ant-based is still a mistery to me.

martindholmes:
@dmj Don't worry about Test2; @sydb and I will take care of that. Although if you want to learn how it works, there's some decent documentation. Try running `ant -projecthelp | less` in the Test2 folder.

sydb:
Right, but I wasn’t actually asking for you to generate a formal test in either Test/ or Test2/, but rather to give me a test file (or point me at what I would do to generate one) just so I can test the PR! :-)
(I guess you can either think of me as being thorough or lazy or both. :-)

dmj:
@sydb You can take https://github.com/dmj/Stylesheets/blob/t/ant-test-to-html/Test2/inputFiles/test-382.xml and the expected result in https://raw.githubusercontent.com/dmj/Stylesheets/t/ant-test-to-html/Test2/expected-results/test-382.html

sydb:
@dmj — I was prodded to look at this again (by the ever-efficient @HelenaSabel), and see that I never posted a reply to your [02 Nov 20 post](https://github.com/TEIC/Stylesheets/pull/475#issuecomment-720256685). Those links give me 404 errors. (And did when I tried them some time ago, but I have no idea how long ago that was. May have been 03 Nov 20 or 03 May 22 for all I know. :-) But more importantly, they still give me 404 errors now.

sydb:
So my instinct is not to let this fix go for an _entire year_ without pulling it in. Especially since it seems pretty straight forward, and the code changes look fine at first glance.
But not only can I not find the test files you refer to, @dmj, I cannot find the entire repo, including the branch to pull in from.

dmj:
Both, test and repo are long gone. I'll try to recreate them.

peterstadler:
Release Team looked at it and approved for current release 7.54.0

PR #477, opened 2020-11-05
Gracefully fail if target if @facs does not exist

### LINKED ISSUE #476: ANT transformations don't work in Oxygen XML Editor with certain document
If I try to publish the XML document from inside [facsimile.zip](https://github.com/TEIC/Stylesheets/files/5494833/facsimile.zip) to EPUB, DOCX or ODT, I get the following error:

`Transformation failed. C:\Users\SORIN_~1\AppData\Local\Temp\build1731161873\copy.xml:1: Use a resource collection to copy directories.`

The problem seems to be related to `<pb facs="#B49r"/>` from inside the document.

Note that the XHTML and PDF transformations run fine.

dmj:


martindholmes (approved):

tools/makegraphicsanttask.xsl
```diff
@@ -91,6 +91,9 @@ of this software, even if advised of the possibility of such damage.
 	     <xsl:when test="tei:match(@rend,'none')"/>
 	     <xsl:when test="starts-with(@facs,'tcp:')"/>
 	     <xsl:when test="starts-with(@facs,'unknown:')"/>
+             <xsl:when test="starts-with(@facs, '#') and empty(id(substring(@facs, 2)))">
```
peterstadler: thank you very much for that fix, but – as most of the time – other issues surface as well ;)
First, I think I'd prefer to capture that case (of invalid references) in the following choose, where those local references are (trying to be) resolved, see https://github.com/TEIC/Stylesheets/blob/f410b156d26dc3b6e845f14faf1681e176e7a6b1/tools/makegraphicsanttask.xsl#L97

The bigger issue I fear is that the script nowhere deals with multiple pointers, e.g. `<pb facs="#page1a #page1b"/>` (which is valid TEI). This would probably entail a more thorough rewrite of this script?

martindholmes:
@peterstadler I wouldn't want to hold back on a bugfix just because there's more that could be done. There's always infinitely more that could be done. :-) But I think your proposal to move the error trap is good.

dmj:
Huh. Guess what: ```tools/makegraphicsanttask.xsl``` is the wrong place to look. Or not the only place. Parts of the code in ```tools/makegraphicsanttask.xsl``` is duplicated in ```epub/epub-common.xsl```.

https://github.com/TEIC/Stylesheets/blob/f410b156d26dc3b6e845f14faf1681e176e7a6b1/epub/epub-common.xsl#L510

This is not good. If you create ODT or DOCX, ```makegraphicsanttask.xsl``` is used. If you create an ePub, ```epub/epub-common.xsl```.

dmj:
Moved the checks inside the xsl:otherwise as requested by @peterstadler, added a check for empty source file name & ported the checks to epub-common.xsl

sydb:
Quick question @dmj, per [your comment](#issuecomment-723440597) above: I agree, this is not good. Should another ticket be opened or is this something you have subsequently addressed in tihs PR?

dmj:
@sydb I did not address this issue in my PR and I don't know what another ticket would track. I consider the Stylesheets as broken beyond repair. They work for a to a certain degree but they are, from my point of view, effectively unmaintainable.

martindholmes (approved):

martindholmes:
I suspect there are projects out there which will get unexpected error messages as a result of this, since they may be relying on post-processing to resolve apparently-bad links, but I think this should be merged. @sydb ?

Issue #471, opened 2020-10-26
Add permalink for examples


hcayless:
See TEIC/TEI#1931.

We want to add a link on each example making it easy to obtain a URL directly to that example. Examples already have IDs, making that straightforward. The links should point at a particular version of the GLs rather than "current".

Also see #470, so that if someone bookmarks one of these links and visits it, they see a message alerting them that they have visited an older version.

martindholmes:
@npcole @sydb and I discussed this, and we think the best approach is:

1. Use the existing ids generated as part of the HTML build process to create a pilcrow anchor.
2. In the processing, check whether the `<egXML>` already has an `@xml:id`, and if it does, use that instead, so that if a policy of adding stable `@xml:id`s to all `<egXMLs>` is instituted, the processing will make appropriate use of them.
3. Shelve the idea of checking whether you're looking at an old version or not; that can best be accomplished by JS checking the access URL, and is nice-to-have but not essential at this stage.
4. Consider instituting the same process for `<eg>`s.

sydb:
The relevant code that generates ¶ links for sections of the _Guidelines_, and thus we would probably want to copy-and-modify or make a subroutine out of, is in guidelines.xsl, starting at line ~678:
~~~xml
      <span class="bookmarklink">
        <a class="bookmarklink" href="#{$ident}">
          <xsl:attribute name="title">
            <xsl:text>link to this section </xsl:text>
          </xsl:attribute>
          <span class="invisible">
            <xsl:text>TEI: </xsl:text>
            <xsl:value-of select="tei:head[1]"/>
          </span>
          <span class="pilcrow">
            <xsl:text>¶</xsl:text>
          </span>
        </a>
      </span>
~~~

martindholmes:
Stylesheets group today implemented this for most examples (prose and spec pages), but we still have to make it happen in the examples pages ("Show all"). We also need to look at egXMLEndHook, where the link text "bibliography" is hard-coded, when it should actually be an i18n call. This will be the subject of a second ticket.

PR #481, opened 2020-12-11
With @martindholmes do <eg> part of #471

martindholmes:
Update which adds ids and anchors to `<eg>` examples in Guidelines pages, to match what already happens with `<egXML>`s.


martindholmes:
The build failure looks like it's just due to diffs -- egs have ids and anchors where they didn't before. If the pull request is accepted, the next stage would be to update the tests.

sydb:
I think we should update the tests before the PR is accepted, no?

JanelleJenstad (approved):
Okay with me!

martindholmes:
@sydb You're right about the tests. Let's do that first.

sydb:
OK, done. There were 4 actual results that were not matching expected results in Test/. I looked at the diff of each, and all differences look appropriate, so I updated the expected_results/.
Note: Boy, am I happy with the new DIFFNOW parameter to Test/Makefile. Made this comparatively easy to do.
All tests (except those with `xelatex`) passing on my local machine, and Mr. Travis seems happy with things now.

sydb:
## Updates

### security CSS change
@martindholmes discovered that using the `@style` attribute for in-line CSS (rather than a `class` attribute to refer to CSS elsewhere) causes problems on some servers due to CORS security restrictions (cross-origin resource sharing). Thus we have changed the `@style` on our new `<a>⚓</a>` to a `@class`. **Warning**: This means that the change here has to be considered along with commit 4bd962e (branch “issue_stylesheet_471_eg_anchors”) in the TEI repo, as that is where the CSS is stored.

### newlines caused ugliness
@npcole, @martindholmes, and I discovered that if an `<eg>` ended in a newline, the “bibliography Show all ⚓” line looks ugly (because it floats too low). We considered various solutions, but decided that

- newlines at the beginning and end of some examples, but not others, makes the example boxes look mildly inconsistent in the _Guidelines_
- better to constrain documents than have “magic” code to strip them in the Stylesheets
- TEI should not restrict the contents of `<eg>` for users, it is just for the _Guidelines_ that newlines should not be at beginning nor end

So we added a constraint (in the “issue_stylesheet_471_eg_anchors” branch of the TEI repo) to the `<eg>` element _in p5odds_ that forbids leading or trailing newline, and updated the _Guidelines_ to match. (If you really want a leading or trailing newline, just put a NO-BREAK SPACE (U+00A0) before the leading newline or after the trailing newline.)

## Upshot

In order to review this PR, you will need to review [TEI repo PR 2079](https://github.com/TEIC/Stylesheets/pull/481#issuecomment-748269928) simultaneously.

PR #485, opened 2021-02-15
only try to number notes inside tei:text for OpenEdition output

rvdb:


martindholmes:
@rvdb I merged this, then it occurred to me to check whether the numbering function also takes account of the constraint to ignore notes in the header:

~~~~~
<xsl:function name="local:get.note.nr" as="xs:integer">
    <xsl:param name="node"/>
    <xsl:choose>
      <!-- Count footnotes continuously throughout the document is this is set in the stylesheet parameter -->
      <xsl:when test="$footnote.number.continuous">
        <xsl:number select="$node" level="any"/>
      </xsl:when>
      <!-- Otherwise, restart footnote numbering for each front, body, or back section -->
      <xsl:otherwise>
        <xsl:number select="$node" level="any" from="tei:front|tei:body|tei:back"/>
      </xsl:otherwise>
    </xsl:choose>
  </xsl:function>
~~~~~

Am I right that the `<when>` clause should have `@from="tei:text"`?

rvdb:
Hi @martindholmes , of course, thanks for checking! I've changed this, which ended up as a separate PR: https://github.com/TEIC/Stylesheets/pull/486.

Issue #488, opened 2021-02-25
constraintSpec in elementSpec/attList/attDef causes borked output


martindholmes:
Following on from this ticket: https://github.com/TEIC/TEI/issues/2115

Where an elementSpec has an attDef which contains a constraintSpec, the constraintSpec is not rendered where it should be (in the attDef context); instead, its content is blurted out in a messy way in the content model of the element itself. This is manifested in the spec pages for variantEncoding, rt, citeStructure and moduleRef in the 4.2.0 release of TEI.

sydb:
I have not found the problem, but it certainly smells like the code that processes the contents of `<attDef>` is applying or calling the template for `<constraintSpec>` in the wrong mode.

martindholmes:
It's in odd2lite.xsl:
~~~~~
java -jar Utilities/lib/saxon9he.jar -s:p5.xml -xsl:../../Stylesheets/odds/odd2lite.xsl -o:testp5lite.xml
~~~~~

And the problem is right there.

The issue shows up when the attDef is processed in the context of the element content model; the constraintSpec should be ignored but it's not.

PR #489, opened 2021-02-26
Fix for #488:

sydb:
Remove constraintSpec elements from attLists prior to tangling for use in generating documentation

### DIFF
--- a/common/common_tagdocs.xsl
+++ b/common/common_tagdocs.xsl
@@ -739,9 +739,9 @@
                 </xsl:otherwise>
               </xsl:choose>
               <!--
-	      <xsl:for-each select="$myatts/a">
-		<xsl:copy-of select="*|text()"/>
-	      </xsl:for-each>
+              <xsl:for-each select="$myatts/a">
+                <xsl:copy-of select="*|text()"/>
+              </xsl:for-each>
 -->
             </xsl:element>
           </xsl:element>
@@ -940,8 +940,20 @@
                   <xsl:call-template name="showClassAtts"/>
                 </xsl:for-each>
               </xsl:if>
-              <xsl:apply-templates mode="tangle"
-                select="../tei:attList"/>
+              <!--
+                  We want the attributes here as the result of
+                  tangling our sibling <attList>(s). However, we do
+                  not want the <constraintSpec>s to be processed, so
+                  here we remove them first, before tangling. See
+                  https://github.com/TEIC/Stylesheets/issues/488
+                  and
+                  https://github.com/TEIC/TEI/issues/2115.
+                  —Syd, 2021-02-25
+              -->
+              <xsl:variable name="attList_sans_constraintSpecs">
+                <xsl:apply-templates select="../tei:attList" mode="nuke_constraintSpec"/>
+              </xsl:variable>
+              <xsl:apply-templates mode="tangle" select="$attList_sans_constraintSpecs"/>
               <xsl:for-each select="..">
                 <xsl:call-template name="defineContent"/>
               </xsl:for-each>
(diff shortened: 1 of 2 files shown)


sydb:
Note for reviewers—there are only 2 blocks of changed code: starting at line 943 and at line 963=975; everything else is just incidental whitespace changes (sorry, @peterstadler). All that said, the more important review is to run it, IMHO. 🙂

martindholmes:
@sydb When I build this, the content model of rt looks a lot nicer, but it's missing the three locally-declared attributes -- am I misunderstanding something?

~~~~~
element rt
{
   att.global.attributes,
   att.global.rendition.attributes,
   att.global.linking.attributes,
   att.global.analytic.attributes,
   att.global.facs.attributes,
   att.global.change.attributes,
   att.global.responsibility.attributes,
   att.global.source.attributes,
   att.typed.attributes,
   att.placement.attributes,
   att.transcriptional.attributes,
   att.editLike.attributes,
   att.written.attributes,
   macro.phraseSeq
}
~~~~~

That means that (e.g.) the conversion element goes from this:

~~~~~
element conversion
{
   att.global.attributes,
   att.global.rendition.attributes,
   att.global.linking.attributes,
   att.global.analytic.attributes,
   att.global.facs.attributes,
   att.global.change.attributes,
   att.global.responsibility.attributes,
   att.global.source.attributes,
   att.datable.attributes,
   att.datable.w3c.attributes,
   att.datable.iso.attributes,
   att.datable.custom.attributes,
   att.formula.attributes,
   attribute fromUnit { teidata.pointer },
   attribute toUnit { teidata.pointer },
   empty
}
~~~~~

to this:

~~~~~
element conversion
{
   att.global.attributes,
   att.global.rendition.attributes,
   att.global.linking.attributes,
   att.global.analytic.attributes,
   att.global.facs.attributes,
   att.global.change.attributes,
   att.global.responsibility.attributes,
   att.global.source.attributes,
   att.datable.attributes,
   att.datable.w3c.attributes,
   att.datable.iso.attributes,
   att.datable.custom.attributes,
   att.formula.attributes,
   empty
}

So I don't think we're there yet. Talk to you in an hour...
~~~~~

martindholmes:
I believe the problem occurs because the context is lost when the attList is spun off into a variable, so when down the line a key() function is used, it fails to find what it's supposed to find because it doesn't have a context to operate on. I think you can make this error explicit by adding the `@as` attribute:

~~~~~
<xsl:variable name="attList_sans_constraintSpecs" as="element(tei:attList)">
   <xsl:apply-templates select="../tei:attList" mode="nuke_constraintSpec"/>
</xsl:variable>
~~~~~

which then triggers this:

~~~~~
TEI/P5/buildweb.xml:2: The following error occurred while executing this line:
TEI/P5/antbuildweb.xml:32: Fatal error during transformation using TEI/P5/Utilities/guidelines.xsl: 
In the key() function, the node supplied in the third argument (or the context node if absent) 
must be in a tree whose root is a document node; 
SystemID: Stylesheets/odds/teiodds.xsl; Line#: 348; Column#: 4
~~~~~

But this gives us a clue to another place we could look at:, in teiodds.xsl line 348:

~~~~~
  <xsl:template match="tei:attRef" mode="tangle">  
    <xsl:choose>
      <xsl:when test="key('IDENTS',@class)">
	<ref xmlns="http://relaxng.org/ns/structure/1.0" name="{tei:generateAttRef(.,$generalPrefix)}"/>
      </xsl:when>
      <xsl:when test="@class"/>
      <xsl:otherwise>
	<ref xmlns="http://relaxng.org/ns/structure/1.0" name="{@name}"/>
      </xsl:otherwise>
    </xsl:choose>
  </xsl:template>
~~~~~

Is there a way that we can provide info to that template so that it can distinguish between the time when it's building the attDef output and the time when it's building the parent element's content model? If so, we could suppress the constraintSpec in the latter case.

sydb:
Righto. Embarrassingly I got the more explicit error message when I had an `@as`, and was pleased and befuddled when it went away when I was no longer asserting what the variable was. Sigh.
So two possible solutions jump to mind:

1. As you suggest, teach that template (likely by sending it a parameter) whether it is being used to generate the content model documentation, or
2. Figure out how to supply a proper context to the key() function.

sydb:
@martinascholger, @npcole, @martindholmes and I are working on solution (1), above, using a tunnel parameter. @martindholmes should be checking in a first attempt shortly.

sydb:
Seems to have worked!

PR #493, opened 2021-03-17
added a default text() processing template that just copies text() no…

rvdb:
…des through, and avoids interference with upstream text() processing in the standard TEI stylesheets.

### DIFF
--- a/profiles/jtei/jtei.common.xsl
+++ b/profiles/jtei/jtei.common.xsl
@@ -340,6 +340,14 @@
     </xsl:choose>
   </xsl:template>
 
+  <!-- Default for text() nodes: just copy unmodified. 
+       (This template avoids interference with upstream
+       text() processing in the standard TEI stylesheets.
+  -->
+  <xsl:template match="text()" priority="-1">
+    <xsl:value-of select="."/>
+  </xsl:template>
+
   <!-- This function checks if a node ends with quotation marks that pull in subsequent punctuation. -->
   <!-- For the context node and all its descendants containing the last non-empty text node, it is
        tested if any of these is a "quotation element". -->


martindholmes:
@rvdb What problem does this address? I haven't seen any errors with jtei odt recently.

rvdb:
@martindholmes There appeared to be a structural problem where leading white spaces were trimmed from text nodes that followed elements. 

You can see this e.g. when you transform the source file of your jTEI article at https://journals.openedition.org/jtei/3106 to ODT, and look for e.g. following sentence: 

    "To avoid anything like this, it made sense to specify that content documents only have content, which is easy to achieve using the @startattribute on <schemaSpec>(example 1)"

The lacking spaces at the end of that sentence went unnoticed because I was debugging with the https://github.com/TEIC/Stylesheets/blob/46350387daf9200468b6e198e4fe10f75f288062/profiles/jtei/odt/odt.common.xsl stylesheet. Yet, the actual transformation is done by https://github.com/TEIC/Stylesheets/blob/46350387daf9200468b6e198e4fe10f75f288062/profiles/jtei/odt/to.xsl, which includes the ODT building routines of the standard TEI stylesheets. In this upstream processing, these `text()` nodes ended up being matched by the template at: https://github.com/TEIC/Stylesheets/blob/46350387daf9200468b6e198e4fe10f75f288062/common/functions.xsl#L497

This resulted in unwanted space-trimming. The ODT transformation is the only jTEI transformation that imports the standard TEI framework, which explains why it hasn't appeared in the other output formats. I've added the extra template here (and not in the importing `odt.common.xsl` stylesheet), to make sure the template at: https://github.com/TEIC/Stylesheets/blob/46350387daf9200468b6e198e4fe10f75f288062/profiles/jtei/jtei.common.xsl#L321 wouldn't risk being overridden. 

So, yes, this extra template most probably only affects these `text()` nodes in the ODT processing, by shielding them from the standard TEI processing; since none of the other jTEI transformations are using the general TEI stylesheets, for those transformations this extra template will just come down to an explication of the default (built-in) processing of `text()` nodes.

Best,

Ron

martindholmes:
Thanks @rvdb! Now I see it.

PR #494, opened 2021-03-17
jTEI set default language

rvdb:


sydb:
@rvdb —
The fix to $jtei.lang looks good. (And is a clever XPath, at that. :-) But I want to check with you that the the text() template override is doing what you want. I have not taken the (hours of) time to figure out _which_ upstream template(s) this would override, but I am a bit worried that it would not because the priority is so low. (Of the 183 templates that have "text()" in their `@match` attrs, 0 have priority < -1, 23 have a priority = -1, and 160 have a priority > -1. Of course most of those are not remotely of interest, but it does make me wonder.)

rvdb:
Thanks for the kudos, @sydb :-). I think Git got me again, by unintendedly (as far as I'm concerned) including that second commit in this PR. It was part of another PR which has been merged already (https://github.com/TEIC/Stylesheets/pull/493). @martindholmes basically made the same remark there, so I definitely should have been more explicit about the rationale for that commit. I have elaborated in  https://github.com/TEIC/Stylesheets/pull/493#issuecomment-801274589 on the "upstream" template that is now overridden by this template. The bottom line is: since there was no default template for `text()` nodes in https://github.com/TEIC/Stylesheets/blob/46350387daf9200468b6e198e4fe10f75f288062/profiles/jtei/jtei.common.xsl, the "upstream" template in the imported stylesheet at https://github.com/TEIC/Stylesheets/blob/46350387daf9200468b6e198e4fe10f75f288062/common/functions.xsl#L497 was producing unwanted side-effects. Adding this default template for `text()` nodes solved this.

But I'm confused (as ever) what the presence of that commit in this PR means after it got merged already in https://github.com/TEIC/Stylesheets/pull/493...

martindholmes:
If the two fixes happened to be made in the same branch, then the second would automatically be included in any outstanding pull request for the first fix. I've been caught by that before. :-)

rvdb:
@martindholmes Well, that's the point: I made sure both were in different branches, but somehow the first commit was included in this branch as well (maybe I should have based the later branch on `dev` again, instead of the earlier branch I used for the first commit). Anyway, I hope this doesn't pose any problems!

sydb:
Okey-dokey, @rvdb.

Issue #487, opened 2021-02-19
HTML5 doctype created by the stylesheets is obsolete


martindholmes:
All HTML5 output from the stylesheets, including the Guidelines, contains the following long-obsolete formula:

~~~~~
<!DOCTYPE html
  SYSTEM "about:legacy-compat">
~~~~~

This was originally proposed as a sort of temporary hack for systems unable to produce the proper HTML5 doctype (https://dev.w3.org/html5/html-author/#doctype-declaration). It's a hindrance in that processors such as Saxon will stumble over it, making post-processing of TEI-generated HTML more difficult, and since producing the proper doctype is now trivial with modern versions of Saxon, we should eliminate this old form.

We should first update all our Saxon jars to make sure we don't have any old versions still floating about in the repos; then attend to all instances like this:

~~~~~
  <xsl:output encoding="utf-8" method="xml" doctype-public="-//W3C//DTD XHTML 1.1//EN"/>
~~~~~
and this:
 ~~~~~
  <xsl:output method="xml" omit-xml-declaration="yes" doctype-system="about:legacy-compat" />
~~~~~

moving instead to modern XHTML5:
~~~~~
<xsl:output method="xhtml" html-version="5" encoding="UTF-8" indent="yes" normalization-form="NFC"
    exclude-result-prefixes="#all" omit-xml-declaration="yes"/>
~~~~~

which will automatically give us the correct HTML5 doctype and properly-namespaced XHTML.

sydb:
Hear hear!

sydb:
@HelenaSabel & @martindholmes — What is left to do on this ticket?
I note that `<!DOCTYPE html
  SYSTEM "about:legacy-compat">` no longer appears in the output of `bin/teitohtml`, but still appears in the HTML version of the _Guidelines_.

martindholmes:
That must be due to the guidelines.xsl.model, surely -- let's take a look this morning when we meet.

HelenaSabel:
The reason is that [PR 499](https://github.com/TEIC/Stylesheets/pull/499) has not been merged yet because I wanted to make some final changes, but it will be over soon and we can close this ticket :)

martindholmes:
@sydb and I just made a change to the xsl:output method in the TEI repo guidelines.xsl.model to make it html5, but that has no effect on the Guidelines build because it's overridden at odds/odd2html.xsl in the Stylesheets. 

[...]

PR #499, opened 2021-04-06
update html output declarations to modern XHTML5 #487

HelenaSabel:
My first attempt to address issue #487 (and my first contribution to the Stylesheets!)

In [`html/html_param.xsl`](https://github.com/TEIC/Stylesheets/blob/876f99a9c85656a518c5feb28a322e669c3af688/html/html_param.xsl#L588-L599) I added the new parameters to be consistent with how the stylesheets were calling the previous parameters related to the the old doctype declaration.

Besides updating the instances of `<xsl:output>` and `<xsl:result-document>` that created HTML documents, I modified [this line](https://github.com/TEIC/Stylesheets/blob/876f99a9c85656a518c5feb28a322e669c3af688/html/html_param.xsl#L335-L337) because otherwise the UTF-8 encoding was being declared twice.


sydb:
Woo-hoo!
Two minor concerns, one big problem:

1. (Minor) Should we be setting `@omit-xml-declaration` to `"yes"`? Seems to me we would, in general, want them. But I am not much of an HTML and browser person, so there may be gotchas I don’t know about; and perhaps it is another ticket, eh?
2. (Minor) Not that it matters, but is there anything wrong with declaring the character set encoding twice?
3. (**Major**) As the output HTML files are changed, this breaks not only the tests in Test/ (kind of expected, they are very fragile) but also those in Test2/ (not surprising, because although they are not nearly as fragile, these tests are still comparing serializations, not trees).

hcayless:
For modern HTML, I would say omitting the XML Declaration is correct. The character encoding thing is sloppy, but probably doesn't break anything. But we should fix it, because it's sloppy. As Syd says the PR should be updated to fix the expected results.

HelenaSabel:
>     1. (Minor) Should we be setting `@omit-xml-declaration` to `"yes"`? Seems to me we would, in general, want them. But I am not much of an HTML and browser person, so there may be gotchas I don’t know about; and perhaps it is another ticket, eh?

To my understanding, the XML declaration is not needed unless the encoding is not UTF-8. I am partial to omitting it, but I need to find more solid arguments than the first one that came to mind (because it was related to having the XML declaration in files with a `.php` extension, which is not the point of this). It seems more practical to address the presence or absence of the declaration in this ticket, but if it ends up needing a long discussion it might be better to divide it in two tickets?

>     2. (Minor) Not that it matters, but is there anything wrong with declaring the character set encoding twice?
So the issue was that that `<xsl:when>` was creating a second `<meta>` element with just the encoding attribute: any browser would render it without problems but I think the repetition is invalid.

>     3. (**Major**) As the output HTML files are changed, this breaks not only the tests in Test/ (kind of expected, they are very fragile) but also those in Test2/ (not surprising, because although they are not nearly as fragile, these tests are still comparing serializations, not trees).
For my personal testing, I manually encoded the expected changes (like adding the HTML doctype declaration) but I didn’t know how to delve into a more sustainable testing process...

martindholmes:
I'm pretty sure the HTML validator will complain if the character encoding is declared twice. I think Saxon 10 provides one character encoding declaration, so if we're using a recent Saxon, we shouldn't have to add anything extra.

HelenaSabel:
I finally finished with the testing so this PR is almost ready to be reviewed. I say almost because I had some doubts about the merging of the HMTL and HTML5 routines, therefore some work might still be needed.

At this point, there are no HTML5-specific stylesheets. Any reference to `html5.xsl` was replaced with `html.xsl` and of course `html.xsl` and the `html` folder were updated accordingly. What I haven’t done, though, was to delete the `html5` folders. For instance, the `html5` folders exist in different profiles, but the `to.xsl` file they contain points to `html.xsl` instead of `html5.xsl` (see [example](https://github.com/TEIC/Stylesheets/blob/754a20242a4984067879541931a4d181fbbe2fef/profiles/tei/html5/to.xsl)). I am guessing that the most elegant solution would have been to delete the HTML5 references and have the `html5` commands to just be aliases of the `html` ones, but I am not sure I can do that on my own.

Regarding the merge in terms of the XSLT transformation itself, I had a doubt about the parameter `divOffset`. This parameter being equal to `1` was present in `html5.xsl` and thus I added it to `html.xsl` (ec6205b). One of the changes you will see now when comparing the new expected results is that all the headings will be consistently one hierarchical step higher. I just wanted to make sure that that was the wanted behavior.

Since there indentation is quite different, the expected results are a bit hard to compare with the previous version. These are the main changes you will find:

- No XML declaration and no other doctype but the HTML doctype declaration: `<!DOCTYPE html>` 
- `h2` > `h1`, `h3` > `h2` and so on as mentioned above due to the addition of the `divOffset` parameter
- Semantic HTML5 elements: `<header>`, `<section>`, `<nav>`...
- Creation of `<div>` elements for `<tei:front>`, `<tei:body>` and `<tei:back>` (as it was part of the HTML5 transformation)
- Use of `@lang` attribute. By reading [this W3C recommendation](https://www.w3.org/International/questions/qa-html-language-declarations), I understand that the attribute `@xml:lang` shouldn’t be used.
- Replacement of deprecated `@mode` attribute in `<math>` with `@display="inline"`

And I think that covers it!

martindholmes:
@HelenaSabel This is great work. It'll be an easy thing to tweak the symlinks in the /bin folder to point the html5 one to the html one, which will then take over, I think.

The only thing I'm not sure about is dropping the xml:lang attribute. Since we don't know whether end-users will choose to serve their documents as XHTML or not, I see no problem in leaving the xml:lang attribute alongside the lang attribute, as the W3C document says; that's what I tend to do in my projects.

JanelleJenstad:
I don’t have an opinion on xml:lang but wanted to second Martin’s accolades. Thanks for all this work!

Janelle Jenstad, Associate Professor, Department of English, University of Victoria<https://journals.uvic.ca/index.php/scene/>
Director, Map of Early Modern London<https://mapoflondon.uvic.ca/>; Director, Linked Early Modern Drama Online<https://lemdo.uvic.ca/>
Co-Coordinating Editor, Digital Renaissance Editions<https://digitalrenaissance.uvic.ca/> and the new Internet Shakespeare Editions
Co-Applicant, The Endings Project<https://endings.uvic.ca/index.html>
Email: ***@***.******@***.***>, ***@***.******@***.***>, or ***@***.******@***.***>

HelenaSabel:
> @HelenaSabel This is great work. It'll be an easy thing to tweak the symlinks in the /bin folder to point the html5 one to the html one, which will then take over, I think.

Thank you @martindholmes! I thought it could be more complicated than that. I could tackle the XSLT dependencies, but I was afraid of some other type of dependencies related to the other programing languages that work here.

> The only thing I'm not sure about is dropping the xml:lang attribute. Since we don't know whether end-users will choose to serve their documents as XHTML or not, I see no problem in leaving the xml:lang attribute alongside the lang attribute, as the W3C document says; that's what I tend to do in my projects.

Thank you for this comment, because I was doubtful myself (in fact, I was adding both at some point a48b53e). I changed my mind at re-reading the recommendation because the presence of both was only mentioned when serving XHTML 1. But you are absolutely right, if the pages are served as XML, `@lang` does not hurt but `@xml:lang` would be crucial. 

I will fix these details tomorrow (first thing in the morning) and put the PR ready for review.

HelenaSabel:
I did the symlink in the `bin` folder. And the `ll` command shows that the symlink has the proper target, but the stylesheets in the `html5` folders inside `profiles` are still being fired. I see three possible approaches: 
1. Keep the XSLT files as they are in folders like `profiles/default/html5` or `profiles/tei/html5`. These stylesheets are calling for the `html.xsl` one (because `html5.xsl` no longer exists) so there is no inconsistency.
2. Create a symlink in all these folders. For instance, `profiles/default/html5/to.xsl` would be a symlink to `profiles/default/html/to.xsl`
3. Update `transformtei` to add some kind of condition for the `html5` command.
 
Option 1 is already implemented and I could do option 2. Sadly, I cannot do option 3.

martindholmes:
@HelenaSabel I'm surprised that the symlink from teitohtml5 to teitohtml fails, but I think a simpler approach might be just to make the profiles/tei/html5 folder a symlink to profiles/tei/html, so the old html5 files no longer exist. Then we could deprecate the teitohtml5 symlink and the html5 folder, and in a year or two, just remove them. What do others think?

HelenaSabel:
Converting to draft to work on the symlinks

html/html_figures.xsl
```diff
@@ -189,7 +189,7 @@ of this software, even if advised of the possibility of such damage.
 	<xsl:apply-templates/>
       </xsl:when>
       <xsl:otherwise>
-	<xsl:element name="{if ($outputTarget='html5') then 'figure'
+        <xsl:element name="{if ($outputTarget=('html5', 'html')) then 'figure'
```
HelenaSabel, line 195: Here I was cautious and just added 'html' to the condition, but I am partial to eliminating the condition. The output targets that use these HTML stylesheets, like epub, shouldn’t have any issues with the element `<figure>`

html/html.xsl
```diff
@@ -71,7 +73,10 @@ of this software, even if advised of the possibility of such damage.
   <xsl:include href="html_textcrit.xsl"/>
   <xsl:include href="html_transcr.xsl"/>
   <xsl:include href="html_verse.xsl"/>
-
+  <xsl:include href="html_textstructure_id.xsl"/>
+  <xsl:param name="divOffset">1</xsl:param>
```
HelenaSabel, line 77: The parameter `divOffset` and its value were taken from `html5/to.xsl`. However, I don’t understand why this was the default for HTML5.

html/html_param.xsl
```diff
@@ -332,7 +332,7 @@ will generate an &lt;h2&gt;</p>
       </xsl:if>
       <meta name="generator" content="Text Encoding Initiative Consortium XSLT stylesheets"/>
       <xsl:choose>
-	<xsl:when test="$outputTarget='html5' or $outputTarget='epub3'">
+	<xsl:when test="$outputTarget='epub3'">
```
HelenaSabel, line 335: For XHTML5, the `<meta/>` element with the `@charset` declaration is already being created thanks to the output declaration

Test/Makefile
```diff
@@ -147,63 +147,65 @@ css:
...
 		else echo "==deferring: \` diff  $(AR)/test4.html $(AR)/test3.html \`"; fi
 
-	$(SAXON) test5.xml ../html/html.xsl  autoBlockQuote=true $(SAXONOPT)  | xmllint --format - > $(AR)/test5.xhtml
+	$(SAXON) test5.xml ../html/html.xsl  autoBlockQuote=true $(SAXONOPT)  | tidy -config config_tidy.txt - > $(AR)/test5.xhtml
 	if [ $(DIFFNOW) -eq 1 ]; \
 		then diff  $(AR)/test5.xhtml $(ER)/test5.xhtml; \
 		else echo "==deferring: \` diff  $(AR)/test5.xhtml $(ER)/test5.xhtml \`"; fi
-	$(JING) -c xhtml.rnc $(AR)/test5.xhtml
+#	$(JING) -c xhtml.rnc $(AR)/test5.xhtml
+# This schema is not compliant with XHTML5 (it doesn’t allow the presence of @lang attribute).
```
sydb: This comment could use some improvement. I _think_ what HBS means is roughly “The xhtml.rnc schema previously used for validation is not actually an XHTML5 schema, but rather an XHTML 1.1 schema. Notably, it does not permit `@lang`, which we need. Thus I have provisionally commented-out the validation against it until we can set up proper XHTML5 validation. — hbs 2021-08-30”.
If I am right, @HelenaSabel, you can just copy that wording in. If I am wrong, a) we’re glad I didn’t just change it myself, and b) strong evidence it needs to be re-worded. 🙂

html/html.xsl
sydb: Me neither.

html/html_figures.xsl
sydb: My suggesting is to leave the code as you have it, and add a comment to the effect of `<!-- Not sure why we feel it necessary to check output type and use <figure> iff HTML ... <figure> should work for anything that calls these HTML stylesheets, methinks. — hsb, 2021-08-30 -->`.

sydb (approved):
Things I have done to review this PR:

- Run bin/teitohtml (which is now same thing as bin/teitohtml5)
-- validated output with W3C validator
-- validated output with oXygen
-- opened output in browser
- eyeballed all changes (a few minor updates recommended in comments below)

Things I have _not_ done include:
- read every frickin’ line of code carefully
- tested epub output
- tested slides

sydb:
Because we are in refrigeration most, if not all, of us would say this PR is technically ineligible to be merged in for the upcoming release. (It is a code change, not a mere wording change.)
For that reason, _I_ am not even going to recommend it be merged now, let alone do so. But if the release techs (other than HSB) agree to it (i.e., if they do not mind taking on the risk), I think they could reasonably merge it before forking the release.

HelenaSabel:
I think it might be better to wait. I will add your suggestions to the PR, @sydb, and I also need to run Test2 (even if I don't find any errors, it is likely that I will need to update the expected results of that test)

Test/Makefile
HelenaSabel: That was exactly what I meant: thank you for the rewording the comment!

sydb:
Fine with me to wait. And yes, there are Test2/ expected result errors.

Test2/build_utilities.xml
```diff
@@ -238,7 +238,7 @@
       Current test processes don't process any such files.
     </description>
     <echo level="info" message="Clean ${inFile} ready for diffing with expected results."/>
-    <replaceregexp file="${inFile}" match="^\s*&lt;!DOCTYPE[^>]+>" replace=""/>
+   <!-- <replaceregexp file="${inFile}" match="^\s*&lt;!DOCTYPE[^>]+>" replace=""/>-->
```
HelenaSabel, line 241: This line was no longer necessary because the old doctype wasn’t being declared, but I had to comment it out because it was generating an encoding issue in the results.

martindholmes: Thanks @HelenaSabel. I'm very happy to see that line disappear. :-)

martindholmes (approved):
Looks good to me!

sydb:
I ran the same tests I ran previously, and also successfully built the _Guidelines_ against this set of stylesheets.

Issue #504, opened 2021-04-30
Missing space after idno in biblStruct processing (obvious in Guidelines bibliography)


martindholmes:
@sydb and I, working on a TEI ticket, noticed that where an idno (eg for a doi) exists in a biblStruct, and its processing is followed by the processing of a journal title, there is no space between them in the output (both in the HTML and PDF Guidelines). Although we believe the simplest solution here is to re-encode the relatively few remaining biblStructs entries as bibls (we had a campaign to convert them a few years ago, but it was never completed), in the meantime we should fix the missing space.

martindholmes:
Working fixes for this are on branch sydb-504_space_after_DOI, and I think we're ready for a pull request. Waiting for @sydb to confirm.

sydb:
On my local machine:
- Quick eyeballing of output looks right.
- All tests in Test2/ pass
- All tests in Test/ pass (except building PDF, which always fails)

So yeah, I guess we are ready to pull & roll.

PR #505, opened 2021-04-30
Address issue #504 by adding a space after idno elements inside biblStruct

sydb:
Making this relatively high priority because in quite a few cases the missing space actually makes it very difficult for a user to parse the DOI from the date, for example.

(Note I do not think this requires all 4 invited reviewers; I think as soon as 1 or 2 of said “OK” the assignee can go ahead and merge.)


martindholmes:
@sydb No takers on this so far. Shall we just go ahead and merge, before things in dev change and it gets more complicated?

sydb:
Hmmm … we (Council) are meeting this weekend. How about I bring it up then to try to move it along?
There. Just added to the agenda. (For North American break-out, Fri 07 @ 19:00Z. Since 3 of the 4 requested reviewers will be there, I think that should be fine.)

martindholmes:
@sydb Sounds good. All this for one leetle waffer-thin space.

JanelleJenstad (approved):
Looks good. Okay to merge.

Issue #241, opened 2017-03-12
A preserveOrder= of "false" does nothing


sydb:
In an ODD, `<sequence preserveOrder="false">` produces the same RELAX NG as `<sequence>` (or `<sequence preserveOrder="true">`).

For testing, try [this tiny ODD](https://github.com/TEIC/Stylesheets/files/836590/Tpo.odd.txt) that tries to constraint `<choice>` to contain 1 `<expan>` and 1 `<abbr>` in either order. Then try it changing "false" to "true". (Note that, to appease GitHub, I have added a “.txt” extension to the filename.)

sydb:
N.B. [TEI issue #1970](https://github.com/TEIC/TEI/issues/1970).

martindholmes:
Pinging this ticket --  surely we need to get this fixed? It's currently not possible (as far as I can see) to create a content model which may contain zero or one of each of a list of elements in any order.

joeytakeda:
I've recently been bitten by this bug and confirm it's still happening.

I've done a bit of digging in the Stylesheets and it looks like the root of the problem (or, at least, the initial problem) is in `odd2odd.xsl`. Running this command against the `dev` branch of the Stylesheets (where preserveOrder.odd is the ODD @sydb provided):

```
saxon -s:preserveOrder.odd -xsl:../Stylesheets/odds/odd2odd.xsl -o:preserveOrder_expanded.odd

```

Shows a `<sequence>` for content that has lost its `@preserveOrder` entirely

```
          <content>
            <sequence minOccurs="1" maxOccurs="1">
              <elementRef key="abbr"/>
              <elementRef key="expan"/>
            </sequence>
          </content>
```

I _think_ the problem is in the `odd2odd-simplifyODD` template and, in particular, this condition:

https://github.com/TEIC/Stylesheets/blob/96001864018d9708ab44523fcafe3475f6885d6f/odds/odd2odd.xsl#L1489-L1493

Only the `@minOccurs` and `@maxOccurs` is retained from the context element and all other attributes are ignored; just did a quick test by adding `@xml:id` to the `<sequence>` and it's also absent from `preserveOrder_expanded.odd`.

I'm happy to work on this in a fork and file a PR, if that would be helpful!

joeytakeda:
The problem was indeed in `odd2odd.xsl` as far as I can tell. `teiodds.xsl` (imported by`odd2relax.xsl`) already has handling for `preserveOrder`:

[...]

PR #507, opened 2021-05-22
Addressing `@preserveOrder` for #241

joeytakeda:
This batch of commits modifies `odd2odd.xsl`, which was dropping `@preserveOrder` (and all other attributes) on `<sequence>`. There are also some new tests added to Test2 to confirm that `preserveOrder` is reflected in the RNG output.


Test2/inputFiles/invalidInstances/testPure1InvalidInstance.xml
```diff
@@ -149,6 +149,12 @@
         should be invalid.</p>
       </div>
       
+      <div>
+        <head>Testing <att>preserveOrder</att> on <gi>sequence</gi></head>
+        <p>This <choice><reg>test</reg><orig>Test</orig></choice> should be invalid, as <gi>reg</gi> ought to appear
```
sydb: Umm, no.
This should be invalid because the `<orig>` ought to appear before the `<reg>`.

odds/odd2odd.xsl
```diff
@@ -1382,6 +1382,7 @@ of this software, even if advised of the possibility of such damage.
     <xsl:variable name="minOmaxO" select="tei:minOmaxO( @minOccurs, @maxOccurs )"/>
     <xsl:variable name="min" select="$minOmaxO[1]"/>
     <xsl:variable name="max" select="$minOmaxO[2]"/>
+    <xsl:variable name="otherAtts" select="@*[not(local-name() = ('minOccurs','maxOccurs'))]"/>
```
sydb: I would much prefer to test the nodes, rather than their names. I.e.
~~~xml
<xsl:variable name="otherAtts" select="@* except ( @minOccurs, @maxOccurs )"/>
~~~

odds/odd2odd.xsl
```diff
@@ -1490,6 +1491,11 @@ of this software, even if advised of the possibility of such damage.
           <xsl:element namespace="http://www.tei-c.org/ns/1.0" name="{$element}">
             <xsl:attribute name="minOccurs" select="$min"/>
             <xsl:attribute name="maxOccurs" select="if ($max eq -1) then 'unbounded' else $max"/>
+            <!-- Copy the attributes declared on the source element -->
+            <xsl:copy-of select="$otherAtts"/>
+            <!--joeytakeda iss241: I'm not sure if the @* below is necessary -->
```
sydb, line 1502: I think not only are you correct, @joeytakeda, you are so correct that in the old days I would have suggested that, rather than add this comment, you should just delete the "@*|" from the `@select` which is now on line 1499. Nowadays, though, there are those among us who think we should _not_ fix _other_ things on any given PR.

sydb (changes requested):
Overall I think this looks excellent. It solves a problem I wish we did not have (I think having `@preserveOrder` with default of "true" on `<sequence>` is a really bad way to do this). But certainly if this is going to work, the attriubtes need to survive. Thus overall I want us to merge in @joeytakeda’s fix, with thanks.
There is, however, 1 logic error in the prose of testPure1InvalidInstance.xml that needs to be fixed first, lest it confuse future readers out of their minds.

joeytakeda:
Thanks, @sydb — agreed on all fronts here. I'll make those changes now.

sydb (approved):
I ran this through both Test2/ and Test/ tests (except the xelatex test, which will not run on my system). Passed them all. This should be merged, IMHO. (Despite the fact that I don’t _like_ this way of handling unordered, but required children.)

sydb:
@joeytakeda For reasons I do not understand, some of the tests in Test/ are now failing. I have both fixed them in a commit to this repo on my local machine, and @HelenaSabel and I have changed the Test/Makefile program so that these should not cause errors in the future. (Simultaneously making the test slightly weaker.)
So could you either
1. Give me r/w permission to this repo, or
2. Merge in the latest from the upstream TEIC/Stylesheets:dev branch into here, and try re-running the tests yourself.
Thanks.

joeytakeda:
Hi @sydb I've merged the changes from TEIC/Stylesheets:dev into the branch and both `make test` and `make test2P` are successful locally (the tests have failed here, but I think that's just because of the Slack webhook?)

sydb:
@joeytakeda — Sorry, did not notice your reply of 9 days ago. (And I even left this tab open in my browser for 2 weeks just so I would see it! Sigh.)

This passes all tests on my local system (except [this bug](#562) interfered), except for whitespace in the test30.rnc output. So I think it should be merged right away, before more divergence occurs. (And we can fix the test30 output after it is merged.)

Oh. It is assigned to me. So I will merge it now, and hope to fix the test30 problem within an hour.

PR #508, opened 2021-06-04
Fix #503 with a <choice> for processing type= of <tag>

### LINKED ISSUE #503: common_tagdocs.xsl ignores `<tag>/@type`
The spec for the `<tag>` element includes values for `@type` including "start, "end", "empty" etc. 

https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-tag.html

However, the template that renders the element, in common_tagdocs.xsl (l.1852) ignores those values:

~~~~~
 <xsl:template match="tei:tag">
    <xsl:element namespace="{$outputNS}" name="{$hiName}">
      <xsl:attribute name="{$rendName}">
        <xsl:text>tag</xsl:text>
      </xsl:attribute>
      <xsl:text>&lt;</xsl:text>
      <xsl:apply-templates/>
      <xsl:text>&gt;</xsl:text>
    </xsl:element>
  </xsl:template>
~~~~~

I think appropriate content should be inserted, so that:

- "end" inserts a slash after the opening angle bracket
- "empty" inserts a slash before the closing bracket
- "comment" inserts "!--" before and "--" after the content
- "pi" inserts question marks before and after the content
- "ms" inserts escaped `[CDATA[` before and `]]` after the content

Otherwise I don't see much point in these `@type` values at all.

sydb:
Passes both `make test` and `(cd Test2 && ant test)` on my local system.


joeytakeda:
This has now been superseded by #618, so I am going to close this (but not the underlying issue)

PR #510, opened 2021-06-10
Look in <idno> as well as <altIdent> for module FPI

### LINKED ISSUE #509: <idno> in addition to, then instead of, <altIdent> for FPI of module
Per TEIC/TEI#2049, we need to change the Stylesheets so that the FPI of a module is indicated by `moduleSpec/idno[@type eq 'FPI']` as opposed to `moduleSpec/altIdent`. Of course, since the latter will be deprecated, not removed, for a year or two, they need to be handle both possibilities for at least that long.

sydb:
Use more complex XPath to take child `<idno>` if there is one, child `<altIdent>` if not, as portion of FPI.

### DIFF
--- a/odds/odd2dtd.xsl
+++ b/odds/odd2dtd.xsl
@@ -206,7 +206,7 @@ of this software, even if advised of the possibility of such damage.
 	&lt;!ENTITY % file.</xsl:text>
             <xsl:value-of select="@ident"/>
             <xsl:text> PUBLIC '-//TEI P5//ELEMENTS </xsl:text>
-            <xsl:value-of select="tei:altIdent[@type='FPI']"/>
+            <xsl:sequence select="( tei:idno[@type eq 'FPI'], tei:altIdent[@type eq 'FPI'] )[1]"/>
             <xsl:text>//EN' '</xsl:text>
             <xsl:value-of select="@ident"/>
             <xsl:text>.dtd' &gt;
@@ -231,7 +231,7 @@ of this software, even if advised of the possibility of such damage.
 	&lt;!ENTITY % file.</xsl:text>
             <xsl:value-of select="@ident"/>
             <xsl:text>-decl PUBLIC '-//TEI P5//ENTITIES </xsl:text>
-            <xsl:value-of select="tei:altIdent[@type='FPI']"/>
+            <xsl:sequence select="( tei:idno[@type eq 'FPI'], tei:altIdent[@type eq 'FPI'] )[1]"/>
             <xsl:text>//EN' '</xsl:text>
             <xsl:value-of select="@ident"/>
             <xsl:text>-decl.dtd' &gt;
@@ -1308,7 +1308,7 @@ of this software, even if advised of the possibility of such damage.
       <xsl:text> </xsl:text>
       <xsl:text>PUBLIC</xsl:text>
       <xsl:text> '-//TEI P5//</xsl:text>
-      <xsl:value-of select="key('FILES',@key)/tei:altIdent[@type='FPI']"/>
+      <xsl:sequence select="( key('FILES',@key)/tei:idno[@type eq 'FPI'], key('FILES',@key)/tei:altIdent[@type eq 'FPI'] )[1]"/>
       <xsl:text>//EN' '</xsl:text>
       <xsl:value-of select="@key"/>
       <xsl:text>.dtd'</xsl:text>
--- a/Test2/expected-results/testNonTeiOdd1.rng
+++ b/Test2/expected-results/testNonTeiOdd1.rng
@@ -36,9 +36,12 @@
                            xmlns:gcse="http://www.google.com/cse"
                            xmlns:sch="http://purl.oclc.org/dsdl/schematron"
                            xmlns:xh="http://www.w3.org/1999/xhtml"
-                           test="if (child::xh:h3) then count(ancestor::xh:div) = 0                             else if (child::xh:h4) then count(ancestor::xh:div) = 1                             else if (child::xh:h5) then count(ancestor::xh:div) = 2                             else not(child::xh:h3 or child::xh:h4 or child::xh:h5)"> The level of your heading (h3, h4 or h5) must fit with the depth of
-                           your division nesting; so the top-level div would have h3, a div inside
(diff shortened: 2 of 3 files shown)


sydb:
Adding @martindholmes to review changes in Test2/.

martindholmes (commented):
Not sure why this module reorganization is necessary for this ticket, but it looks OK. @sydb Why the refactoring into separate modules?

sydb:
Just to have a place to put `<altIdent>` and `<idno>`! :-)

martindholmes:
Ah, I see! No impact at all on expected-results from this?

sydb:
Right, none other than those to Test2/expected-results/testNonTeiOdd1.rng, which are entirely from the whitespace commit, not the actual change.

sydb:
Made this high priority, because it is blocking most further work on TEIC/TEI#2049.

martindholmes:
I'm happy with it, but I think someone else from Council should approve too.

sydb:
Agreed.

sydb:
BTW, @hcayless, @peterstadler, and anyone else looking at this closely, you mostly want to look at the changes to odds/odd2dtd.xsl. The other changes are to a test file and its RNG output which a) is not as vital, and b) @martindholmes has already looked at.
If you do look at the ODD and RNG files, I deliberately checked in whitespace changes separately to try to make comparing the real changes easier. Of course, _I_ don’t know how to do that comparison here on GitHub. Care to enlighten me?

odds/odd2dtd.xsl
```diff
@@ -206,7 +206,7 @@ of this software, even if advised of the possibility of such damage.
 	&lt;!ENTITY % file.</xsl:text>
             <xsl:value-of select="@ident"/>
             <xsl:text> PUBLIC '-//TEI P5//ELEMENTS </xsl:text>
-            <xsl:value-of select="tei:altIdent[@type='FPI']"/>
+            <xsl:sequence select="( tei:idno[@type eq 'FPI'], tei:altIdent[@type eq 'FPI'] )[1]"/>
```
peterstadler, line 209: just curious: why did you change the `xsl:value-of` to `xsl:sequence`? The latter is returning a node and you're relying on the output method to finally return the atomic value. Why not stick with `xsl:value-of`  to immediately return a string?

peterstadler (approved):
All in all, this PR is quite small (I expected much worse) and easy to read, yet I needed to do my own research to find the corresponding ticket and discussion …
The changes do favour `<idno>` over `<altIdent>` which seems to be intended.

odds/odd2dtd.xsl
sydb: Because I was told the exact opposite — that `<value-of>` returns a text node that needs to get converted to a string. We can check into it at our leisure, one instance is not going to make any significant difference.

martindholmes: I think @sydb is right: value-of constructs a text node, whereas sequence  doesn't have that overhead:

https://stackoverflow.com/questions/43755377/xslsequence-vs-xslvalue-of-in-function-which-is-better-to-use

sydb: Yeah, `<value-of>` wraps its result in a text node whether it is needed or not; `<sequence>` avoids this extra overhead. See [an explanation by M. Kay](https://stackoverflow.com/questions/43755377/xslsequence-vs-xslvalue-of-in-function-which-is-better-to-use). In this case I bet it makes no difference, since the things before and after are text nodes.

peterstadler: @martindholmes , @sydb but the issue is that we want text nodes here, don't we? (at least, that would be in line with the rest of the output of that template). And I think it's not about performance here, but about clarity. 
If you wanted to stick with  `xsl:sequence`, the `@select` should include an explicit type cast/atomization IMHO, like 
```xml
<xsl:sequence select="( tei:idno[@type eq 'FPI'], tei:altIdent[@type eq 'FPI'] )[1]/string()"/>
```

sydb: So first off, I think it is just so cool that @martindholmes and I responded with pretty close to the same answer in the same minute even though we are a > 5,000 km drive away from each other!

I don’t have a strong opinion on this (I want the PR in so I can move ahead with other aspects of `<altIdent>`, and there is so much problematic with the Stylesheets writ large, this tiny issue really doesn’t bother me either way). That said, perhaps whole thing really should be something like
~~~xml
    <xsl:if test="@type ne 'core'">
      <xsl:sequence select='
          "&#x0A;&lt;![%TEI."
        ||normalize-space( @ident )
        ||";[&#x0A;&#x20;&lt;!ENTITY % file."
        ||normalize-space( @ident )
        ||" PUBLIC &apos;-//TEI P5//ELEMENTS "
        ||( tei:idno[@type eq "FPI"], tei:altIdent[@type eq "FPI"] )[1]
        ||"//EN&apos;&#x20;&apos;"
        ||normalize-space( @ident )
        ||".dtd&apos;&#x20;>&#x0A;&#x20;%file."
        ||normalize-space( @ident )
        ||";&#x0A;&#x20;]]>&#x0A;"
        '/>
    </xsl:if>
~~~

martindholmes: @peterstadler @sydb This has forced me to realize that there's something I hadn't quite understood before: A DTD is not an XML file. The output method for odd2dtd is text. So my question is whether we do want any text "nodes" here; my default assumption would be that you can't have a node unless you have a tree, and you can't have a tree unless you have XML output. But my understanding of DTDs is rudimentary at best.

sydb: Well, in one sense a DTD is, of course, an XML file since its syntax and semantics are defined by the XML specification. But it is not in the XML instance syntax which we all know and love so well. (I believe there were a couple of efforts at re-expressing DTDs in an XML instance syntax, but don’t have time to look for them now.)

But for the moment, I do not really care whether it is 1‰ better (in some sense of the word “better”) to generate a text node or not. I would really like to proceed with `<altIdent>`; IMHO if we want to discuss this issue more we should do so on the list or Slack, as it is not really germane to only this one particular instance of text generation (or any other particular instance).

PR #514, opened 2021-06-25
Issue 399 rng annotations

martindholmes:
This is a simple fix for the exact issue reported. Instead of dropping the contents of the code element completely, as was happening before, it outputs them as text surrounded by backticks, which is (we think) a reasonable way to signal that something is inline code in a plain-text context.


sydb (approved):
I just tried generating output from the test ODD using this branch, and it seems to have worked just fine.

PR #517, opened 2021-08-06
Fix #516: duplication of ids in HTML transformation

### LINKED ISSUE #516: Repetition of @id when processing listBibl/bibl into HTML
While testing the XHTML5 conversion I got a validation error that was already present in the expected results: the repetition of the same `@id` values. This is caused when processing [test32.xml](https://github.com/TEIC/Stylesheets/blob/dev/Test/test32.xml) which contains `listBibl/bibl` elements with `@xml:id` attributes.  In the HTML output, the value of the `@xml:id` is added both to the `<li>` element and to its child `<div>`: See line [4639](https://github.com/TEIC/Stylesheets/blob/5bdca2bfc328ed1a66f595687d60a7fb27c72ce1/Test/expected-results/test32.html#L4639) and following. 

The culprit is [here](https://github.com/TEIC/Stylesheets/blob/5bdca2bfc328ed1a66f595687d60a7fb27c72ce1/html/html_core.xsl#L846) and it _should_ be a very straightforward fix. The doubt I have concerns the output we want: should we generate a `<li>` element with or without the `<div>` child? I would say without it, but there might be some consistency issues to consider...

HelenaSabel:
This PR tackles issue #516: In the transformation to HTML, the `@xml:id` of `listBibl/bibl` elements was being added twice: first in the `<li>` element and then in its `<div>` child. 

My proposal is to just add a condition: If the child of `listBibl` being processed has already an `@xml:id`, do not call the `makeAnchor` template.

I am assigning as reviewers @ebeshero and @sydb because of your participation in the discussion of the original ticket.

### DIFF
--- a/Test/expected-results/test32.html
+++ b/Test/expected-results/test32.html
@@ -4636,19 +4636,19 @@ color:orange;
         <span class="head">Bibliografija</span>
       </h2>
       <ol class="listBibl">
-        <li id="Kovacic">
+        <li>
           <div class="biblfree" id="Kovacic">KOVAČIČ – Fran Kovačič: Služabnik božji Anton Martin Slomšek, knezoškof lavantinski. Družba sv. Mohorja, Celje, I. 1934, II. 1935. </div>
         </li>
-        <li id="Kramberger">
+        <li>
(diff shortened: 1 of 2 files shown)


ebeshero (approved):
@HelenaSabel This looks good to me--glad we're producing valid HTML in the expected-results! :-)

sydb (approved):

PR #521, opened 2021-08-21
Fix for #520

### LINKED ISSUE #520: links from examples to BIB failing
I have **not** yet investigated what is going on, but at first blush we have a serious problem: `@source` links from `<egXML>`s in the GLs to the bibliography are not working. First glance says it is because the ID of each `<bibl>` is not getting through to the BIB.html file. (Thus I am suspicious of @HelenaSabel’s recent ID fix which, IIRC, I approved just yesterday.)

Mildly more detailed:

- Built GLs in branch attlocatable-issue-1910, found “bibliography” links from examples were not working (took me to BIB.html file, but not proper entry); return links from BIB to example worked.
- Looked in BIB.html, did not see `@id` on entries.
- Built GLs in branch dev, found same (also discovered, kinda accidentally, that many of the “bibliography” links are simply missing, but that is a [different ticket](https://github.com/TEIC/TEI/issues/1476)).

HelenaSabel:
Syd’s fix for #520: it adds processing of `@xml:id` to the application of templates from `<biblStruct>` or `<bibl> `child of `<listBibl>`.

### DIFF
--- a/odds/guidelines.xsl
+++ b/odds/guidelines.xsl
@@ -1,5 +1,17 @@
 <?xml version="1.0" encoding="utf-8"?>
-<xsl:stylesheet xmlns="http://www.w3.org/1999/xhtml" xmlns:a="http://relaxng.org/ns/compatibility/annotations/1.0" xmlns:html="http://www.w3.org/1999/xhtml" xmlns:rng="http://relaxng.org/ns/structure/1.0" xmlns:tei="http://www.tei-c.org/ns/1.0" xmlns:teix="http://www.tei-c.org/ns/Examples" xmlns:xhtml="http://www.w3.org/1999/xhtml" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:xsl="http://www.w3.org/1999/XSL/Transform" exclude-result-prefixes="xlink rng tei teix xhtml a html xs xsl" version="2.0">
+<xsl:stylesheet
+    xmlns="http://www.w3.org/1999/xhtml"
+    xmlns:a="http://relaxng.org/ns/compatibility/annotations/1.0"
+    xmlns:html="http://www.w3.org/1999/xhtml"
+    xmlns:rng="http://relaxng.org/ns/structure/1.0"
+    xmlns:tei="http://www.tei-c.org/ns/1.0"
+    xmlns:teix="http://www.tei-c.org/ns/Examples"
+    xmlns:xhtml="http://www.w3.org/1999/xhtml"
+    xmlns:xlink="http://www.w3.org/1999/xlink"
+    xmlns:xs="http://www.w3.org/2001/XMLSchema"
+    xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+    exclude-result-prefixes="xlink rng tei teix xhtml a html xs xsl"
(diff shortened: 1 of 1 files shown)


sydb:
IMHO:
- This PR **must** be reviewed by someone other than @sydb and @HelenaSabel, and **must** be merged prior to release.
- Once a reviewer says go ahead, I think anybody can merge it in pronto, it does not require both reviewers to OK.

martindholmes (commented):
This looks OK to me, but I'm struggling to figure out what the original $id variable was supposed to be doing:

`<xsl:variable name="id" select="@xml:id"/>`

As far as I can see, it was never used, so something was already broken here. Do you see any use for it?

martinascholger:
I can confirm that the links from examples to BIB work in my local build. But there is no tooltip anymore (maybe only in my local version?).

sydb:
Futhermucker. Do you think lack of tool-tips is just high priority, or outright release blocking?

HelenaSabel:
Same thing happens in my local installation as well, @martinascholger .
I took a quick look at @martindholmes JavaScript and I don’t see the problem: the parameter that it is being sent to the function `showPopupBibl()` matches an `@id` in BIBL.html. I don’t know why it defaults to linking to the bibliography (without showing the popup). I will take a closer look tonight.

martindholmes:
This is normal browser behaviour. AJAX won't run on you local machine, only when the page is coming from a server. It's not broken.:)

sydb:
> … I'm struggling to figure out what the original $id variable was supposed to be doing:
> `<xsl:variable name="id" select="@xml:id"/>`
> As far as I can see, it was never used, so something was already broken here. Do you see any use for it?

I saw so little use for it, I deleted it. :grin:

odds/guidelines.xsl
```diff
@@ -695,8 +707,7 @@
   <!-- link from bibl back to egXML -->
   <xsl:template
       match="tei:listBibl/tei:biblStruct|tei:listBibl/tei:bibl">
-    <xsl:apply-templates/>
-    <xsl:variable name="id" select="@xml:id"/>
+    <xsl:apply-templates select="@xml:id | node()"/>
```
peterstadler, line 710: where is the `@xml:id` transformed to (html) `@id`?

sydb: Short, slightly facetious answer is “I have no idea”. But the fact that I don’t know is the main reason I wanted to go with `@xml:id | node()` rather than `@* | node()` — I can test that wherever `@xml:id` is getting matched it is working. I can’t easily test _all_ attributes.

HelenaSabel: > where is the `@xml:id` transformed to (html) `@id`?

https://github.com/TEIC/Stylesheets/blob/3d9eb7fcc053bd3d74d903949aa067a7771fab2a/html/html.xsl#L874

peterstadler:
Thanks to @sydb and @HelenaSabel trying to explain the issue and the fix during todays Stylesheets meeting! It involves importing this Stylesheet  into the Guidelines' build (https://github.com/TEIC/TEI/blob/59a2becc4a35f4a50a2518c605e1adef310e9f2c/P5/Utilities/guidelines.xsl.model#L20-L21) and some more intricacies.  @sydb put it so: "We had to concede, in the end, that we are not entirely sure how this ever worked before these changes"

PR #524, opened 2021-09-24
overridability for docxtotei

eduarddrenth:
docxtotei offers a construct based on style names starting with "TEI " or "tei_".

This PR enables developers to write threir own elementFromStyle template for example to solve style naming problems or to add custom markup.

See: https://bitbucket.org/fryske-akademy/jurisprudentia-frisica/src/master/customDocxToTei.xsl

### DIFF
--- a/docx/from/docxtotei.xsl
+++ b/docx/from/docxtotei.xsl
@@ -38,7 +38,6 @@
 	  <xsl:import href="omml2mml.xsl"/>
 	  <xsl:import href="pass0.xsl"/>
 	  <xsl:import href="pass2.xsl"/>
-
 	  
 	  <xsl:param name="convertGraphics">true</xsl:param>	  
 	  <xsl:param name="mathMethod">mml</xsl:param>	  
--- a/docx/from/docxtotei.xsl
+++ b/docx/from/docxtotei.xsl
@@ -38,8 +38,7 @@
 	  <xsl:import href="omml2mml.xsl"/>
 	  <xsl:import href="pass0.xsl"/>
 	  <xsl:import href="pass2.xsl"/>
-    <xsl:import href="textruns.xsl"/>
-  
+
 	  
 	  <xsl:param name="convertGraphics">true</xsl:param>	  
 	  <xsl:param name="mathMethod">mml</xsl:param>	  
@@ -55,8 +54,8 @@
 	  <xsl:param name="verbose">false</xsl:param>	  
 	  <xsl:param name="processChangeInformation">false</xsl:param>
 	  <xsl:param name="pageHeight">890</xsl:param>
-	  <xsl:param name="pageWidth">576</xsl:param>
-
+      <xsl:param name="pageWidth">576</xsl:param>
+  
 
 	  <xsl:include href="fields.xsl"/>
 	  <xsl:include href="toc.xsl"/>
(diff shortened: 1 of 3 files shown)


bwbohl:
I really like this because it makes the customisation of the `processTextrun` template much easier. In my current work context, I had a similar situation where it was good to provide a custom prefix for the docx style templates. In my case, I introuced my own template `processTextrun`, which would match my prefix and otherwise `xsl:next-match`:

```xml
<xsl:template name="processTextrun">
    <xsl:variable name="style">
        <xsl:value-of select="w:rPr/w:rStyle/@w:val"/>
    </xsl:variable>
    <xsl:choose>
        <xsl:when test="starts-with($style,'CUSTOM ')">
            <!-- do something -->
        </xsl:when>
        <xsl:otherwise>
            <xsl:next-match/>
        </xsl:otherwise>
    </xsl:choose>
</xsl:template>
```

sydb:
Ack! Thank you, @bwbohl. I had not realized (or forgot?) that I was supposed to review this.
Will try to take a more careful look over the weekend.
@eduarddrenth (or @HelenaSabel) — feel free to poke me if I have not said anything by middle of next week.

eduarddrenth:
Great, some activity here....perhaps also look at #586...

sydb (approved):
While I cannot claim to entirely understand the changes made, nothing looks crazy.

What I have done:
- merged this PR into a copy of TEIC/Stylesheets:dev
- built TEI P5 using the merged stylesheets
- ran both Test and Test2 in that merged Stylesheets/ directory
- ran odttotei from both the regular dev branch and from the merged directory on each of a dozen .odt files and compared the results

I found no errors, and all 12 of the .tei output files from the merged directory matched the output files from the dev branch character-for-character.

What I have not done:
- Actually test trying to override from.xsl by specifying my own docxtoTEI.

docx/from/docxtotei.xsl
```diff
@@ -54,8 +53,7 @@
 	  <xsl:param name="verbose">false</xsl:param>	  
 	  <xsl:param name="processChangeInformation">false</xsl:param>
 	  <xsl:param name="pageHeight">890</xsl:param>
-	  <xsl:param name="pageWidth">576</xsl:param>
-
+      <xsl:param name="pageWidth">576</xsl:param>
```
peterstadler, line 56: Within this file there seem to be only whitespace changes not related to the improvement.

peterstadler (commented):
I was only looking at the code and wondering why you wouldn't just overwrite `processTextrun` (as proposed by @bwbohl) but need another named template?  
I assume that in practice you'd need to introduce some conditional expressions (to filter out particular styles) which would require a fundamental rewrite of `elementFromStyle` rather than a (simple) addition to `processTextrun`. But maybe I'm missing something here?

docx/from/docxtotei.xsl
eduarddrenth: `elementFromStyle` concentrates logic in one method and eliminates duplicate code from the original `processTextrun` thus making overriding simpler.As I understand @bwbohl appreciates this pr as a better alternative to overriding the complete `processTextrun`.

peterstadler: Yes, you're right, and I really appreciate the improvement. Many thanks for the reply!

peterstadler (approved):

PR #525, opened 2021-10-02
Attempt to fix TEIC/TEI#2070:

sydb:
Insert `doc/tei-p5-doc/` between the version number and the filename ("VERSION") in the URL that points at the VERSION file for a given version.


ebeshero:
@sydb I've just run the tests on your branch now, and see no build errors! I also poked through the built Readmes (several of them, not every single one) and the links to specs and issues are good. I'm going to merge this now for the good of the release.

Issue #511, opened 2021-06-24
tei-to-markdown.xsl needs a template to handle images


martindholmes:
The Stylesheets Working Group, looking at issue #444, found that the tei-to-markdown.xsl lib has no template for tei:graphic, so graphic elements are being output as nonsense. This should be a simple template to add.

martindholmes:
I'm presuming that we're aiming to implement [CommonMark](https://spec.commonmark.org/), rather than any of the various other flavours. Our task is to go from a TEI `<graphic>` element to this:

`![alt text](thing.png "title text")`

and the main question is where we get the alt text/title text. They can perfectly well be identical, but for valid HTML5 we'll need at least the alt text. I think the place to look first would be first in a parent `<figure>` element: if there's `<figDesc>` and `<head>` there, use `<figDesc>` for the title and `<head>` for the alt; if there's only one, use it for both; and if neither, or no parent `<figure>` element, then check for a `<desc>` child of the `<graphic>` element (yes, that exists), or as a last resort, `@n` on the `<graphic>` element itself.

PR #528, opened 2021-10-21
First pass at issue #511: graphics to markdown images.

martindholmes:
I've added handling for TEI graphic elements in tei-to-markdown.xsl per issue #511. You can test this by running this in the Stylesheets repo root:

`bin/teitomarkdown Test2/inputFiles/generalP5.xml Test2/outputFiles/test.md`

then looking at the file `Test2/outputFiles/test.md`

### DIFF
--- a/markdown/tei-to-markdown.xsl
+++ b/markdown/tei-to-markdown.xsl
@@ -250,14 +250,16 @@ of this software, even if advised of the possibility of such damage.
         <xsl:text>}</xsl:text>
       </xsl:if>
     </xsl:variable>
-    <xsl:text>![</xsl:text>
-    <xsl:sequence select="$altText"/>
-    <xsl:text>](</xsl:text>
-    <xsl:sequence select="xs:string(@url)"/>
-    <xsl:text> "</xsl:text>
-    <xsl:sequence select="$titleText"/>
-    <xsl:text>")</xsl:text>
-    <xsl:sequence select="string-join($dim, '')"/>
+    <xsl:sequence select="
+      '!['
+      ||$altText
+      ||']('
+      ||@url
+      ||' &quot;'
+      ||$titleText
+      ||'&quot;)'
+      ||string-join( $dim, '')
+      "/>
   </xsl:template>
   
   <!-- Suppress all attributes on graphic element. -->
--- a/markdown/tei-to-markdown.xsl
+++ b/markdown/tei-to-markdown.xsl
@@ -2,6 +2,7 @@
     version="2.0"
     xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
     xmlns:tei="http://www.tei-c.org/ns/1.0"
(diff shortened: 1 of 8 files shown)


markdown/tei-to-markdown.xsl
```diff
@@ -223,6 +224,45 @@ of this software, even if advised of the possibility of such damage.
 	 </xsl:otherwise>
       </xsl:choose>
   </xsl:template>
+  
+  <!-- TEI graphic elements convert to a simple structure, but it's
+          complicated figuring out where best to get the alt and title 
+          text from. -->
+  <xsl:template match="tei:graphic[@url]">
+    <xsl:variable name="altText" as="xs:string" select="if (parent::tei:figure/tei:figDesc) then
```
sydb: Seems to me this variable could be set using the more compact and more idiomatic Xpath 
~~~
(
  parent::tei:figure/tei:figDesc[1],
  parent::tei:figure/tei:head[1],
  child::tei:desc[1],
  'graphic'
)[1]
~~~
no? (And you could tack `=> xs:string()` onto the end just to re-enforce that it is a string, but since there is an `@as`, I do not think that’s necessary.) Same for $titleText.

sydb (commented):
My instinct is that this PR is probably fine, but I am hesitant to approve it because I do not understand the syntax used for `@width` and `@height`. Not part of the [CommonMark  specification](https://spec.commonmark.org/0.30/#images), is it?

martindholmes:
@sydb Feel free to simplify the XPath. I wasn't firing on all cylinders when I wrote the if-then-else stuff. Re the width/height: I was looking for a way to handle those attributes because the TEI has them, and most of us would use them; Common Markdown doesn't provide a way to do that, but the widely-used Kramdown editor implements what I've included. The alternative is to output an HTML img tag; we could branch on whether those atts are there, and fall back to an img tag if they are, and that would work for almost all Markdown implementations, but it's not Markdown, of course.

martindholmes:
@sydb I've updated the XPath per your suggestion, and added a missing colon for compliance with the Kramdown syntax.

sydb:
Excellent. I had never heard of [Kramdown](https://kramdown.gettalong.org/syntax.html). Learned something new. (Can I go home now?)

I have run the test above, and it works for me.

Would you be opposed to changing lines 253–260 to the following?
~~~xml
    <xsl:sequence select="
        '!['
      ||$altText
      ||']('
      ||@url
      ||' &quot;'
      ||$titleText
      ||'&quot;)'
      ||string-join( $dim, '')
      "/>
~~~
Either way, I think PR is fine to be merged. Yes, there is still a lot to be done in this conversion, but it is 1 step better than what we have now.

martindholmes:
@sydb Done and tested again.

sydb:
Seems pretty simple, so just merged.

Issue #515, opened 2021-07-02
Unwanted word "Attributes" appears in HTML output tagdocs


martindholmes:
@sydb and I, looking at issue #506, discovered that in HTML output documentation from user ODD files (not the Guidelines), the word "Attributes" appears in the tagdocs tables for elements and attribute classes, not only in the left column as a label, but also in the right column next to it, before the list of attributes. This results from two calls, on lines 2678 and 2696 of common_tagdocs.xsl. The first call mostly happens in elementSpecs, and the second mostly in attribute classes, although that's not 100% the case. If we could figure out what the $autoGlobal variable does, that would help; it seems to be 'false' by default, and that's causing these things to be output, so perhaps somewhere it should be set to 'true' for this specific output purpose.

martindholmes:
A look at the history of the Exemplars in TEI releases suggests that this goes back well over a decade.

martindholmes:
Here's a genuine example of this:

https://hcmc.uvic.ca/wendat/documentation/tei_back.html#TEI.handShift
(Now: https://hcmc.uvic.ca/wendat/mss/documentation/tei_back.html#TEI.handShift — 2022-04-17)

martindholmes:
Working with @HelenaSabel, @npcole, @sydb, and @martinascholger on this in branch issue-515-caption.

npcole:
Do it

sydb:
@martindholmes, @HelenaSabel, @npcole, @martinascholger, and I just met and fixed this puppy. (We hope.)

PR #531, opened 2021-11-25
Issue #515: “Attribute” repeated in tagdoc

martindholmes:
Working with @helenasabel, @sydb, @martinascholger and @nccole on issue #515: rewrote attribute listings output to suppress unwanted word.


sydb (approved):

sydb:
Passed all tests (after @HelenaSabel updated test15, that is) in my local system.

Issue #529, opened 2021-11-14
Schematron namespace prefix generation


sydb:
By my count there are 224 uses of elements in the Schemtron namespace in the PDF output of the _Guidelines_. Of those, 90 are expressed with `<s:*` and 134 are expressed with `<sch:*`.
Clearly this is problematic. We should be using a consistent namespace prefix except in the case where we are demonstrating “Look, you can use whatever prefix you want. You can say `matron:` if you like.”
A quick look says that most, if not all, of the s: prefixes are on Schematron code used inside an example of `<constraint>`; in the tagdocs both the s: and sch: prefixes appear frequently. I have not noticed any other prefixes used for the Schematron namespace.
I am not sure what is going on here, but my first instinct (which may well be wrong) is that at times the source has a prefix already present (sch:), in which case it is used, and at times the source just has `xmlns="http://purl.oclc.org/dsdl/schematron"`, in which case the Stylesheets assign a prefix (s:). This is just a guess.

PR #532, opened 2021-12-13
Per #529, change prefix for Schematron namespaces

sydb:
While trying to explain the problem described in #529 to @martindholmes, we found what we think is the source of said problem, and thus the solution. In common/verbatim.xsl we changed the generated prefix for the ISO Schematron namespace from `s:` to `sch:`; and, correspondingly, changed the generated prefix for the Schematron 1.x namespace from `sch:` to `sch_1.x:`. (It is no longer ever used in the _Guidelines_ per TEIC/TEI#537 I don’t think. So it probably could be deleted, but just in case …)


HelenaSabel (approved):
Indeed leaving the outdated version might seem overcautious, but I agree is the right call. 
All tests passed so it’s ready to merge!

sydb:
Per Council mtg today, merge!

Issue #526, opened 2021-10-07
phrase-level references to macros should be links in PDF


sydb:
In the _Guidelines_ references to macros are, generally, encoded with `<ident type="macro">`.[1] These should (IMHO) produce links in the output HTML (and ePub and whatever) to the tagdoc for the encoded macro. 

#### Notes
[1] I think they should be encoded with `<name type="macro">`, myself. But they are not. Sigh.

sydb:
There are only 3 occurrences of `<ident type=macro>` in the _Guidelines_: in `DEFCONTEI`, `MDMDCM`, and `ppf-err`. The references to macros that are generated by the Stylesheets for the “Schema Declaration” section of the tagdocs are proper links.

martinascholger:
`<ident type="datatype">` should be a link as well

martindholmes:
Waiting on Jenkins to see that all is well. Also we need to check the PDF build and the ePub and Kindle.

martindholmes:
This needs to stay open, because links are not created in the PDF. They are working in the ePub, and therefore presumably in the Kindle version, but I can't check the latter. Changed the title of the issue to reflect the fact that only the PDF presents a problem now.

martindholmes:
For the record: what is currently p. 16 of the Guidelines PDF, section 1.4.1, contains a list of macros, all of which should be links to their respective pages, and all of which are links in the HTML, but they aren't in the PDF, That's what we need to fix.  Oddly enough, links do appear in section 1.5.

martindholmes:
In fact this is a broader problem with the PDF. If you look at p. xviii and search for "References to classes are also presented as links, for example model.divLike for a model class, and att.global for an attribute class", you can see two more things which should be links but are not: model classes and attribute classes. In the input TEI Lite, these appear as `<ident>` elements too. I think the problem lies in latex/latex_core.xsl, in the template at line 337, which simply styles any ident instead of turning it into a link.

martindholmes:
Fix created with @sydb and @npcole in branch issue-526-mdh-pdf-links. Pull request coming.

martindholmes:
This can be closed following a pull request. A follow-up, issue #537, has been raised to cover the slightly different cases of ref elements rather than idents, in e.g. section 1.4.

PR #536, opened 2022-01-03
Issue 526 mdh pdf links

martindholmes:
Proposed fix for missing links for macros etc. per issue #526.

### DIFF
--- a/latex/latex_core.xsl
+++ b/latex/latex_core.xsl
@@ -338,7 +338,6 @@ of this software, even if advised of the possibility of such damage.
     <xsl:text>\textsf{</xsl:text>
     <xsl:choose>
       <xsl:when test="@type = ('macro', 'datatype', 'model')">
-        <!--<xsl:message>PROCESSING IDENT OF RELEVANT TYPE</xsl:message>-->
         <xsl:call-template name="makeInternalLink">
           <xsl:with-param name="dest" select="text()"/>
           <xsl:with-param name="body" select="text()"/>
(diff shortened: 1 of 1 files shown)


sydb:
Nope.
I built the PDF version using this branch (`issue-526-mdh-pdf-links`) of the Stylesheets against the `dev` branch of the Guidelines using Hugh’s Docker container. The PDF still does not have the links you (@martindholmes) specifically mention on the ticket. I believe the problem is that the new line 340 compares against "macro", "datatype", and "model", but not "**class**".

sydb:
Nope.
I went ahead and corrected line 340. (It had "model" where it should have had "class".) But that did not solve the problem. I added debugging code to be sure: the "makeInternalLink" template is being called properly now, but (at least in xviii case) the classes are still not links.
I have removed my debugging code and pushed the fix to line 340 in 8d8ac5f2.
Will spend a few more minutes on this, but then off to bed.

martindholmes:
@sydb are you doing the full pdf build or just the pdfonce?

sydb:
Using `make pdf`, which I think does both `pdfonce` and `pdfrest`.

Found that the `<ident type="class">model.divLike</>` was (reasonably) testing for `id('model.divLike')`, but there **is no** such ID. There is "TEI.model.divLike", which looks like the right thing. So I separated out the test for "class" and prepended “TEI.” to the $dest. Seems to be working, but I am too tired to test carefully.
Updated in a6e34de6.

sydb:
Yes, looks like `make pdf` does `Guidelines.pdf:` (which does `pdfonce`) and `pdf-complete:` (which does `pdfrest`).

sydb:
Note that the headwords (as it were) in “1.4.1 Standard Content Models” on PDF page 16 are still not links. They are encoded as `<ref>`s, not `<ident>`s.
Should we be doing something there, too?

martindholmes:
@sydb did the touch-ups, and we tested with @npcole, so this is ready to merge. A new ticket for the instances of ref rather than ident will be raised.

PR #545, opened 2022-02-15
Issue 541: fix bug when processing <constraintSpec>

### LINKED ISSUE #541: Unable to properly delete a `<constraintSpec>`
One would think this would be easy. Just `<constraintSpec module="[module]" ident="[ident]" sceme="schematron" mode="delete"/>` should do it, eh? A bit weird that `@scheme` is required, but no big deal.

No such luck. The big deal is that it does _not_ work. Nor does most any other method I tried. According to my little [test ODD](https://github.com/TEIC/TEI/files/7919470/KC.odd.txt) (which is attached to this ticket with an extra .txt extension), one can delete a `<constraintSpec>` from a _class_ by specifying the deletion or change (to an empty `<constraintSpec>` or `<constraint>`) in a `<classSpec mode="change">`; and one can delete a `<constraintSpec>` from an _element_ only if `"replace"`ed with an empty one.

I do not think this is the right state of affairs. I think that the whole point of being in att.identifiable is to be deleted or over-ridden.

HelenaSabel:
See issue #541 for the original problem. 

The proposed solution depends on the value of `@ident` attributes inside of `<constraintSpec>` elements being unique (see [TEI issue #2224](https://github.com/TEIC/TEI/pull/2224)): the value of the `@ident` attribute is being used as the key of  `<constraintSpec>` (see pertinent changes in file [odds/odd2odd.xsl](https://github.com/TEIC/Stylesheets/compare/issue-541?expand=1#diff-35c5c4ba81d3cda628ca4b7d8d24a379de3ec0d9b315b19c78df9f763cb9a438)). 

This PR includes the required update of the p5subset (which means that @martinascholger doesn’t need to do it this month).


sydb (approved):
I

1. Ran the tests in Stylesheets/Test/
2. Ran the tests in Stylesheets/Test2/
3. Built a complete copy of the _Guidelines_, including HTML, PDF, and schemas
4. Compared the output of (3) above to my previous build.

As far as I can tell with those tests, there is nothing broken in the build, and this PR is ready to go.

Issue #140, opened 2016-02-04
Create test ODD that gives classRef a workout


hcayless:
I'm not confident that the attributes which can be used to modify `<classRef>`, namely `@expand`, `@include`, and `@except` have been implemented. We need an ODD + test that exercises these.

See TEIC/TEI#1412

lb42:
Added testClass.odd to Stylesheets/Test : the good news is that almost everything appearently works as it should when generating Relax or XSD: there is one minor glitch (test7a)  Predictably the DTD generated is rubbish, of course.

lb42:
(I havent added anything to the Makefile to ensure that this new test file gets exercised during the build though, nor have I added a test XML file; pending review)

JanelleJenstad:
Assigning to @HelenaSabel because this ticket is the test for the #139

PR #549, opened 2022-04-30
Test @include and @except

HelenaSabel:
This PR adds the test mentioned in issue #140 to the test suite. It includes:
- A slight modification to the ODD created by @lb42 : [Test/testClass.odd](https://github.com/TEIC/Stylesheets/blob/issue-140-except/Test/testClass.odd)
- The addition of the test to `Makefile` (with the required modification of the expected results).


Test/Makefile
```diff
@@ -753,8 +753,16 @@ test-namespaces: actual
 	perl -p -i -e 's/generated from ODD source .*//' $(AR)/test35.rnc
 	perl -i -pe 'BEGIN{undef $$/;} s/# ?Schema[^#]+#[^#]+#[^#]+#[^#]+#\n//smg' $(AR)/test35.rnc
 	if [ $(DIFFNOW) -eq 1 ]; \
-		then diff -bBw $(AR)/test35.rnc $(ER)/test35.rnc; \
-		else echo "==deferring: \` diff -bBw $(AR)/test35.rnc $(ER)/test35.rnc \`"; fi
+	then diff -bBw $(AR)/test35.rnc $(ER)/test35.rnc; \
+	else echo "==deferring: \` diff -bBw $(AR)/test35.rnc $(ER)/test35.rnc \`"; fi
```
sydb: Indentation should probably be restored. 😊

Test/testClass.odd
```diff
@@ -45,15 +45,25 @@
...
                <classes>
                   <memberOf key="model.divPart"/>
                </classes>
                <content>
                   <classRef key="testClass"/>
                </content>
             </elementSpec>
+            
+            <elementSpec ident="testP1b">
+               <desc>Alternate: should generate x | y | z</desc>
```
sydb: May as well match the attribute value, no? I.e. “Alternation:”.

Test/Makefile
```diff
@@ -753,8 +753,16 @@ test-namespaces: actual
...
+	
+#HBS 2022-04-30 tests attributes @include and @except
+	$(BINDIR)/teitornc $(FLAGS) testClass.odd $(AR)/testClass.rnc
+	perl -p -i -e 's/generated from ODD source .*//' $(AR)/testClass.rnc
+	perl -i -pe 'BEGIN{undef $$/;} s/# ?Schema[^#]+#[^#]+#[^#]+#[^#]+#\n//smg' $(AR)/testClass.rnc
```
sydb, line 762: I realize this regexp fix was just copied from another file (test21, test15, test30 all use that same icky regex), and I do _not_ think we should do anything different here in this PR. But we should move this from here into cleanup.pl, (on a different issue), no?

sydb (approved):

peterstadler:
Just a reminder to @HelenaSabel to react to @sydb's questions and proposals ;)

Test/Makefile
HelenaSabel: Good catch! I don’t even how I changed the indentation of the lines preceding my addition...

Test/Makefile
HelenaSabel: created issue #580

HelenaSabel:
Thank you for the kind poke, @peterstadler. I think I addressed all of @sydb’s comments.

Issue #544, opened 2022-02-14
add `@classpathref` to `common/teianttasks.xml`


bwbohl:
I'm using the TEI Stylesheets as submodule in some non TEI schema development repositories for building the schemata from my ODD definitions. The setup has a project specific ant build.xml that calls the `dist` target `relaxng/build-to.xml`. As my main repo has its own lib folder with anything needed for building I would like to just submit my classpath-ref to the TEI-anttasks instead of setting up the environment at submodule level. This fails though, because e.g. the `macrodef`called `buildgeneric` relies on saxon being on the JAVA class path. As my repositories are designed to be self sufficient (pulling all build required dependencies, instead of relying on them being on the host JAVA class path) this fails.
A easy fix could be provided by adding a `@classpathref` to the `ant:xslt` calls defined in `teianttasks.xml` that references the `ant:path` defined in: 

 https://github.com/TEIC/Stylesheets/blob/1616df8470d0805356d6dc2b4926c81bec37a554/common/teianttasks.xml#L59-L61

e.g.:

```xml
  <macrodef name="buildgeneric">
    <attribute name="in"/>
    <attribute name="out"/>
    <attribute name="xsl"/>
    <attribute name="type"/>
    <sequential>
      <echo level="info">XSLT generate @{type} files</echo>
      <xslt force="yes" style="@{xsl}" in="@{in}" out="@{out}" classpathref="classpath">
        <xmlcatalog>
          <catalogpath>
            <pathelement location="$catalogfile"/>
          </catalogpath>
        </xmlcatalog>
        <factory name="net.sf.saxon.TransformerFactoryImpl">
          <attribute name="http://saxon.sf.net/feature/xinclude-aware" value="true"/>
        </factory>
      <!-- [ omitted for illustrative purposes] -->
      </xslt>
    </sequential>
  </macrodef>
```

This would allow me to submit my project-classpath as ref when calling the build, e.g.:

```xml
<ant dir="submodules/Stylesheets/relaxng/" antfile="build-to.xml" target="dist" usenativebasedir="true">
            <property name="inputFile" value="${basedir}/${customization.path}"/>
            <property name="outputFile" value="${basedir}/schemata/${odd.basename}.rng"/>
            <property name="defaultSource" value="${basedir}/${dist.dir}/kb_canonicalized.xml"/>
            <property name="verbose">true</property>
            <reference refid="kb.classpath" torefid="classpath"/>
        </ant>
```

[...]

PR #550, opened 2022-04-30
Adding @classpathref to all XSLT calls

HelenaSabel:
PR for issue #544. 

I’ve just tested this branch, and it should be noted that the  author of the original issue (@bwbohl ) has tested it as well so I think it can be merged without requiring additional reviews.


ebeshero:
In case a reviewer is wanted, I just read it all and am glad it was such a simple implementation. `@classpathref` seems like it could be useful in other scenarios requiring dependencies, too.

sydb:
I do not really understand the problem or the solution, but 4 others (@bwbohl, @martindholmes, @HelenaSabel, and @ebeshero) all seem to be happy with this, and it has not broken anything. So, I think it can be merged toot sweet. (I just checked it out and merged in latest from dev, there was nothing to merge.)

bwbohl:
@sydb, thanks for merging! Just my two cents on it: with the `classpathref` the TEI Stylesheets can be called from external scripts that will submit the classpath location. So basically, this is about increasing reusability.

PR #563, opened 2022-07-22
fix processing of namespaces 

### LINKED ISSUE #424: `@ns` in ODD customizations causes horrible problems in RNG

peterstadler:
this is to address and fix #424


raffazizzi:
@peterstadler we made some adjustments at the Stylesheets meeting today and we would like to merge this PR. Can you remove the "draft" label?

peterstadler:
Just removed the draft label as requested by Council members @raffazizzi and @trishaoconnor. NB, I haven't looked at the code anymore but only pressed the button. Sorry for the delay!

sydb (approved):
I do not actually remember this issue very well, but the code here looks reasonable at a glance, and passes (after merging from dev) both Test/ and Test2/ (in the docker) and (again in the docker) a complete build of TEI Guidelines (including HTML in every language we support) works against this branch.

PR #575, opened 2022-09-21
added workaround for a Lodel bug that distorts the string "&lt;/?obje…

rvdb:
…ct" in HTML rendering of jTEI articles

### DIFF
--- a/profiles/jtei/openedition/to.xsl
+++ b/profiles/jtei/openedition/to.xsl
@@ -586,7 +586,7 @@
       <xsl:copy-of select="@rend|node()" copy-namespaces="no"/>
     </q>
   </xsl:template>
-  
+    
   <!-- untag <cit> in paragraph: just process contents -->
   <xsl:template match="tei:cit[ancestor::tei:note]">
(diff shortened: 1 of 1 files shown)


peterstadler:
That looks like a proper workaround, I'm just curious whether that should be reported to https://github.com/OpenEdition/lodel? Since from your description it seems the issue is on their side?

rvdb:
@peterstadler I had indeed checked with the Lodel assistance team, only via email. They confirmed it was a bug in the Lodel rendering code, for which they offered a manual ad-hoc intervention from their side after publication of the article. Since a new version of Lodel is being developed, they won't invest in a more robust solution for this version of Lodel. In order to provide a more structural workaround for the jTEI publication workflow for the time being, I have added this (dirty, but limited and effective) fix to the XSLT that generates the OpenEdition files for input in Lodel.

All to say that it is indeed a workaround for a Lodel bug, the only trace of which is in my email conversation. But thanks for pointing out that Github repo; I wasn't even aware of it! For future Lodel problems, I'll post an issue there first.

PR #583, opened 2022-10-21
Update p5subset

peterstadler:
this PR updates the p5subset.xml used for the Stylesheet tests and overwrites the expected results where necessary.


sydb (approved):
I do not know why dozens of definitions of model classes generated as _alternation, _sequenceOptional, _sequenceOptionalRepeatable, and _sequenceRepeatable are occuring in the newly generated expected results, but not in the original expected results. (It can’t really be because of a change in the subset, can it? Maybe there is some code somewhere that says “if one is used generate all”?)
That said, I do not see it as a problem, just a mystery.

PR #585, opened 2022-10-27
Processing attRefs for attribute list on element spec pages. Fixes #419

### LINKED ISSUE #419: Is <attRef> processed in HTML output?
For solving issue https://github.com/TEIC/TEI/issues/1867, the `@subtype` attribute has been added to att.textCritical via `<attRef>` which is a temporary workaround until it is possible to override the definition of `@type` with a mode=\"replace\". See issue #370. 

@sydb and I now discovered that `@subtype` is in the schema, but the HTML output of the `<lem>`, `<rdg>`, and `<rdgGrp>` specs does not list `@subtype` (e.g. https://tei-c.org/Vault/P5/3.6.0/doc/tei-p5-doc/en/html/ref-lem.html). We assume that the `<attRef>` element is not processed (accordingly)?

raffazizzi:


sydb (commented):
First, congrats to @raffazizzi for some impressive and, moreover, impressively fast detective work on exactly where the problems were.

I generated HTML output with this branch, and compared it to the HTML output of the dev branch. With exception of metadata stuff (primarily timestamps), the only differences were in the tagdoc pages for entryFree, entry, lem, rdgGrp, rdg, and superEntry.

That’s the good news. I examined the pairs of tagdoc pages (dev & fix-419) for entry and for rdg to each other. Correctly, the only differences are in the attributes listed (and the date generated). More good news.

Minor problem was that in entry the following space was missing. That’s not a big deal.

But problematically (IMHO), the new output implies that `<entry>` and `<rdg>` are members of the att.typed class. They are not. I think this is a problem we need to think about.

raffazizzi:
Thanks for this feedback, @sydb! The plot thickens: `lem` (which was the element I looked at for testing) is member of `att.textCritical` just as much as `rdg` is: they should both get `@subtype` from `att.typed` because `att.textCritical` uses `attRef` to cherry-pick it. So the stylesheet is doing exactly what I told it to do, as far as I can tell.

In short:
`lem` and `rdg` are both members of `att.textCritical`, which cherry-picks with `<attRef>` `@subtype` from `att.typed`.

**But** actually the latest `tei_all.rng` does not allow `@subtype` on either of them. `@subtype` does not make it into the RNG definition of `att.textCritical.attributes`. So it looks like we have a new problem, right?

sydb:
No, I _think_ that (RNG does not have `@subtype`) is an old problem, which already has a ticket. (Will look for ticket # later.)

So while the Stylesheets may be doing exactly what you want (which, as I said, I am quite impressed with), I am not sure it is what we want them to do. Do we want the output of a cherry-pick like this to look the same as a `<memberOf>` and `"change"` (or `"replace"`)? Maybe so. But I am not sure. Give me a night to sleep on it, as it were.

raffazizzi:
We may want to find a way to distinguish inheritance via membership from inheritance via cherry-picking, somehow, but _where_ the cherry-picked attribute comes from is important and should be shown.

In the spec pages for attribute classes, this is already shown; e.g. in att.textCritical you can already see:

> [att.written](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-att.written.html) (`@hand`) [att.typed](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-att.typed.html) (~~type~~, `@subtype`)

So I reproduced the same in element spec pages. I would say that for now this is fine so that this information is made available. We can then discuss how to distinguish these two forms of inheritance.

E.g. the new display on lem is:

For membership inheritance (NB lem is not a direct member of att.global.rendition)
 > [att.global](http://localhost:3000/ref-att.global.html) (`@xml:id`, `@n`, `@xml:lang`, `@xml:base`, `@xml:space`) ([att.global.rendition](http://localhost:3000/ref-att.global.rendition.html) (`@rend`, `@style`, `@rendition`))

For both membership and cherry-picked:
> [att.textCritical]() (`@type`, `@cause`, `@varSeq`, `@require`) ([att.typed]() (~~type~~, `@subtype`)) ([att.written]() (`@hand`))

peterstadler (changes requested):
Impressive, indeed!
As @sydb I built the HTML Guidelines and diffed the results. I can confirm @sydb's findings that the tagdoc pages for entryFree, entry, lem, rdgGrp, rdg, and superEntry are properly changed. Yet, the space is missing, which is minor, but should be fixed.

I think this is an improvement and the distinction between cherry-picked and overwritten inheritance is a minor one. I find it more relevant that the spec pages tell you what's in the schema and where it comes from, not how. So, after adding the whitespace this is ready to be merged, I'd say.

raffazizzi:
@peterstadler the space problem should be fixed! Please double check when you get a chance and please merge if it looks good!

joeytakeda:
Merging per Stylesheets mtg 2022-01-05

PR #587, opened 2022-12-02
XLSX→TEI drops empty cells, the PR

### LINKED ISSUE #323: conversion of xslx omitts empty cells at the beginning of a row
The oxgarage conversion process (http://www.tei-c.org/oxgarage/#) from XLSX-spreadsheet to TEI P5 omitts empty cells leading a row:

[emptycells.xlsx](https://github.com/TEIC/Stylesheets/files/2025851/emptycells.xlsx)

where row[2] and row[3] start with an empyt cell.
Result is

```
<TEI xmlns="http://www.tei-c.org/ns/1.0"><teiHeader><fileDesc><titleStmt><title/><author>N.N.</author></titleStmt><publicationStmt><p>No publication statement</p></publicationStmt><sourceDesc><p>A TEI file automatically converted from a XSLX file.</p></sourceDesc></fileDesc></teiHeader><text><body>
<table><head>Table1</head>
  <row n="1"><cell>content1</cell><cell>Case1: online leading cells</cell></row>
  <row n="2"><cell>content2</cell><cell>case 2: leading cell empty</cell></row>
  <row n="3"><cell>content3</cell><cell/><cell>content4</cell><cell>case 3: leading cell empty, empty cell between content cells</cell></row>
</table>
</body></text></TEI>
```

where row[2] and row[3] start with content-cells from column B.

sydb:
I merely implemented the fix @GVogeler outlined on [the ticket](https://github.com/TEIC/Stylesheets/issues/323). Seems to work. Fix did the right thing with the test data he attached to that ticket, and both Test and Test2 pass. However, I do not think there is anything in either test process that actually exercises XLSX→TEI. Sigh.

### DIFF
--- a/debian-tei-xsl/debian/changelog
+++ b/debian-tei-xsl/debian/changelog
@@ -1,3 +1,9 @@
+tei-xsl (7.55.0a) natty; urgency=low
+
+  * new release from upstream
+
+ -- TEI                        <editors@www.tei-c.org>  Thu, 01 Dec 2022 10:33:08 -0500
+
 tei-xsl (7.54.0a) natty; urgency=low
 
   * new release from upstream
--- a/xlsx/xlsxtotei.xsl
+++ b/xlsx/xlsxtotei.xsl
@@ -155,7 +155,7 @@
       <xsl:for-each select="$sheet-document/sml:worksheet/sml:sheetData/sml:row">
         <row n="{position()}">
           <xsl:for-each select="sml:c">
-            <xsl:if test="preceding-sibling::sml:c">
+            <xsl:if test="position() eq 1  or  preceding-sibling::sml:c">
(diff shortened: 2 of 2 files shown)


peterstadler (changes requested):
The Debian changelog file was probably committed by accident. Could you please undo this change to remedy the conflict?

sydb:
In principle I am more than happy to. In practice, I do not know how off the top of my head, and will have to fight with git for a bit …

HelenaSabel (approved):
I think I fixed the conflict so I would say that the PR is ready to merge.

`Test` does include a XLSX to TEI transformation (which didn’t break with these changes). I wonder if a third sheet should be added to `Test/test.xlsx` with some empty cells.

peterstadler (approved):

PR #588, opened 2022-12-12
aligned processing of bibl items with identical author(s) with common…

rvdb:
… code in jtei.common.xsl

This fixes erroneous "----,." output in ODT, for subsequent bibliographic entries with multiple identical authors. After this fix, the output is "----." (without the comma).

### DIFF
--- a/profiles/jtei/odt/odt.common.xsl
+++ b/profiles/jtei/odt/odt.common.xsl
@@ -670,35 +670,15 @@
         <xsl:apply-templates/>
     </xsl:template>
     
-    <xsl:template match="listBibl/bibl">
-        <text:p text:style-name="teiBiblioEntry">
-<!--          Let's see whether this bibl has the same 
-              authors as the previous one. -->
-            <xsl:variable name="authorOrEditor" select="child::*[1]/local-name()"/>
-            <xsl:choose>
-              <xsl:when test="$authorOrEditor = ('author', 'editor', 'orgName') and preceding-sibling::bibl">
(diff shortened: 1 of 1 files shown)


sydb:
I cannot (easily) test this because jtei processing (in particular, Stylesheets/profiles/jtei/odt/odt.common.xsl) uses an EXPath extension function. That requires Saxon PE or EE, and I only run HE locally. The oXygen editor could do it, of course, but I do not know off the top of my head how to ask oXygen to do so.

rvdb:
@sydb Turned out the (single) reference to this extension function had been [commented out](https://github.com/TEIC/Stylesheets/blob/dev/profiles/jtei/odt/odt.common.xsl#L777), so I've commented out the actual function, too. Should work now with Saxon HE.

sydb:
Okey-dokey @rvdb. Thanks.
(But why not just delete it? It is in a version control system, after all. :-)

sydb (approved):
I have now run both Test/ and Test2/ over this branch of the Stylesheets, AND used them to convert a JTEI document that had 2 bibliographic entries that had the same author ([The Count](https://en.wikipedia.org/wiki/Count_von_Count), if you must know) into ODT. All tests passed, and the ODT output had a nice long dash-line without a comma for the 2nd occurrence of the author’s name.

I think this is good to go.

rvdb:
@sydb Nice, the incremental Count :-)! You're right, of course. I was a bit hesitant, since this was @martindholmes 's code, but as it clearly was an abandoned path, I've removed the commented-out stuff now.

PR #589, opened 2022-12-15
Fix for attdef to relaxng (issue #557)

### LINKED ISSUE #557: minOccurs=0 fails to work on `<attDef>`
The `@prefix` attributes of `<schemaSpec>`, `<elementSpec>`, and `<moduleRef>` are defined the same way:

**`<schemaSpec>`**:
~~~xml
      <datatype minOccurs="0">
        <dataRef key="teidata.xmlName"/>
      </datatype>
~~~
**`<elementSpec>`**:
~~~xml
<datatype minOccurs="0"><dataRef key="teidata.xmlName"/></datatype>
~~~
**`<moduleRef>`**:
~~~xml
      <datatype minOccurs="0" maxOccurs="1"><dataRef key="teidata.xmlName"/></datatype>
~~~
In all three cases, the attribute generated is `attribute prefix { xsd:NCName }?` (although in the `<moduleRef>` case it is surrounded by parenthesis, which I do not understand, as the RELAX NG XML syntax looks the same). That seems to me to be a problem, because the definition clearly says that the value can be empty (i.e., contain zero teidata.xmlNames), but the schema says it cannot.

What makes this **really** weird is that in all other cases where an attribute is defined using a `@minOccurs` of '0'[1] the RELAX NG is correct, e.g.: `attribute otherLangs { list { teidata.language* } }?`. OH! But I have just discovered a major difference, evident in the footnote: the ones that work have a `@maxOccurs` of "unbounded".

**Notes**
[1] The 10 attributes defined with `@minOccurs` of '0' are:
~~~
      2 in except, max=unbounded
      2 in include, max=unbounded
      2 in prefix, max=
      1 in atts, max=unbounded
      1 in otherLangs, max=unbounded
      1 in prefix, max=1
      1 in type, max=unbounded
~~~

raffazizzi:
Added case for processing attributes with minoccur 0 and maxoccur 1 as optional in rng. Fixes #557

### DIFF
--- a/odds/teiodds.xsl
+++ b/odds/teiodds.xsl
@@ -1493,7 +1493,13 @@ select="$makeDecls"/></xsl:message>
       <xsl:variable name="min" select="$minmax[1]"/>
       <xsl:variable name="max" select="$minmax[2]"/>
       <xsl:choose>
-        <xsl:when test="tei:datatype/rng:text  or  not( tei:datatype )	or  $max eq 1">
+        <xsl:when test="$min eq 0 and $max eq 1">
+          <optional xmlns="http://relaxng.org/ns/structure/1.0">
+            <xsl:call-template name="attributeData"/>
+          </optional>
+        </xsl:when>
+        <xsl:when test="tei:datatype/rng:text  or  not( tei:datatype ) or $max eq 1">
+          
           <!-- If there is only going to be one output RELAX NG node   --> 
           <!-- in the attribute definition, then we don't need to      -->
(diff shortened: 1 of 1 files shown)


sydb (approved):
Gave correct output when run over the little test ODD I put in the ticket, and passed all tests in Test/ and Test2/ (except the xelatex test, which always fails on my system).

HelenaSabel (approved):
I think it’s ready to merge, but I wonder if maybe we should add @sydb ODD to the test suite?

PR #595, opened 2023-03-05
Update jsi-tei.css

u44026105:


sydb (approved):
I do not know who @u44026105 is, and I don’t _really_ know what JSI is or what this profile is for (but I do know enough that I think @TomazErjavec is the person to ask). It seems the only 2 people who have ever touched this code before were Sebastian Rahtz and @lb42.

All that said, it seems not only pretty harmless, but probably a good idea to add some visual separation between `<bibl>s` in a `<listBibl>`. So my instinct is to just merge this in.

ebeshero:
As @sydb points out, borders are useful, so we disapprove the removal of this in our CSS.

PR #608, opened 2023-05-01
Fix #607, I hope: remove erroneous bits of text from version URL 

### LINKED ISSUE #607: Incorrect URL in comment in output schemas
While working on TEI issue #2417 @sydb discovered an incorrect URL is being generated in the "Edition Location" that is a comment in every output schema. We think the problem is with the variable `$TEIVersionWithoutFullStop` in file teiodds.xsl.

sydb:
With @trishaoconnor, change the "makeTEIVersion" template in odds/teiodds.xsl so that
 a) it does not output the string “P5 Version&#x0A;” (or “Version&#x0A;”) as part of the version URL; and
 b) it outputs the version information on 1 line instead of two.

The good news is that I am quite confident that the new code works, in that it generates the expected output. The bad news is I have no idea if the expected output (since it is slightly different from the previous output) might break something that the build process did not test. (I have fixed the actual Test procedure, and Test2 passed without any problem.)

Warning to reviewers: I have also changed a whole bunch of whitespace, so be sure to ask GitHub to hide whitespace changes when you examine the changes. And I daresay the only file you really want to examine is teiodds.xsl. (All other changes are just fallout from changes to that file.)

### DIFF
--- a/Test/Makefile
+++ b/Test/Makefile
@@ -751,11 +751,11 @@ test-namespaces: actual
 
 	$(BINDIR)/teitornc $(FLAGS) test35.odd $(AR)/test35.rnc
 	perl -p -i -e 's/generated from ODD source .*//' $(AR)/test35.rnc
-	perl -i -pe 'BEGIN{undef $$/;} s/# ?Schema[^#]+#[^#]+#[^#]+#[^#]+#\n//smg' $(AR)/test35.rnc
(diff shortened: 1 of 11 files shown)


jamescummings (approved):
I've not tested it directly, not had time ATM, merely read through the changed files. But it all looks good to me. I suggest another reviewer test it out.

martindholmes (approved):
Both sets of tests pass. I don't pretend to understand what was being done in the Makefile, but everything else makes sense.

PR #615, opened 2023-05-27
Make sure VERSION file is present before reading it.

sydb:
Honestly, I am not entirely sure why we are reading a file rather than having the version number hard-coded in, say, common/common_params.xsl. But in any case, I have:

 * Updated tei:stylesheetVersion() with a fix based on @eduarddrenth’s suggestion.
 * Removed parameter from that function defintion, it was not being used.
 * Removed parameter from the 1 call to that function.
 * Some whitespace changes, too.
 * Fixed my e-mail addr in AUTHORS.
 * Updated expected-results/ to match mildly different comment at the end.

I am afraid I cannot assign @eduarddrenth as a reviewer.


sabineseifert:
Talking about e-mail addresses in AUTHORS: Shouldn't we delete Sebastian's e-mail address in line 3?

sydb:
Good point. Probably should re-write that whole first para. Perhaps something like
~~~
The main author of these XSL stylesheets was
the late Sebastian Rahtz from the University
of Oxford IT Services. The TEI Consortium is
forever in his debt for this invaluable con-
tribution.
~~~

PR #617, opened 2023-07-30
process `@xml:id` wherever the output of a <list> is generated

### LINKED ISSUE #616: xml:id on list element does not give a HTML id
For multilingual quotations I list-item-cit-quote-p (a few items). If I give an xml:id to each of these five elements, the ul element in HTML gets no id. See ul class="horiz" without an id, line 76 in the code of page https://d-meeus.be/marxisme/classiques/MEideall-I.html. I did test an xml:id to the five elements
           list xml:id="MEideall-list2003p7" rend="horiz"
              item xml:id="MEideall-item2003p7" style="width: 48%;"
                cit xml:id="MEideall-cit2003p7" xml:lang="de"
                  quote xml:id="MEideall-quote2003p7"
                    p xml:id="MEideall2003p7" rend="alinea"
(line 13727 in the TEI XML source https://d-meeus.be/marxisme/classiques/lectures.xml)
where only the first fails to pass to HTML. These xml:ids are of course not all needed, but one could like to point to the list as a whole.

sydb:
I am hoping this fixes #616 …
Relatively minor update to html/html_core.xsl: Just added a call to process the `@xml:id` attribute to each of the possible output elements generated for `<list>`. Just chose one when two elements were generated (despite my comment on the ticket : -).
My main surprise is that the tests all seemed to pass on my system. I would have expected diff errors when a test input file had a `<list xml:id="DUCK">` whose output used to be `<ul>` but is now `<ul id="DUCK">`. Harumph.

### DIFF
--- a/html/html_core.xsl
+++ b/html/html_core.xsl
@@ -649,6 +649,7 @@ of this software, even if advised of the possibility of such damage.
       <xsl:when test="@type='catalogue'">
         <p>
           <dl>
+	    <xsl:apply-templates select="@xml:id"/>
             <xsl:call-template name="makeRendition">
               <xsl:with-param name="default">false</xsl:with-param>
             </xsl:call-template>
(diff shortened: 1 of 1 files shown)


tuurma (approved):

tuurma:
What do you mean "just chose one when two elements were generated"? Like, nested elements, e.g. `p/dl` or `p/table` in which case it seems the only sensible option.

sabineseifert (approved):

sydb:
Yes, @tuurma, that is what I think I meant.

PR #618, opened 2023-08-24
Adding templates for val and tag to fix #567

### LINKED ISSUE #567: val and tag elements are ignored
The `<val>` element is ignored by teitohtml.

That is, the input
~~~xml
The <att>n</att> attribute may be specified as <val>0</val>, but only on a full moon.
~~~
is converted to
~~~html
The <span class="att">n</span> attribute may be specified as 0, but only on a full moon.
~~~
rather than
~~~html
The <span class="att">n</span> attribute may be specified as <span class="val">"0"</span>, but only on a full moon.
~~~
which (I claim) is what should be generated.

joeytakeda:
* Template added to html/html_core.xsl for tag and val to transform them into spans
* Named template added to html/html_core.xsl to create a delimiter span to allow easier customization (i.e. to allow for a CSS only approach to start and end text)
* Reflect value of `@scheme` in tag's class

(See https://github.com/TEIC/Stylesheets/issues/567#issuecomment-1692355943 for additional details)

### DIFF
--- a/Test2/build.xml
+++ b/Test2/build.xml
@@ -105,7 +105,7 @@
       without special parameters. -->
     <fileset id="simpleTeiForHtmlFiles" dir="${inputDir}" includes="testP5ExTcp1.xml, 
         testRendition1.xml, testQuotations1.xml, testWhitespace1.xml, testBinaryObjects1.xml,
-        testCorpus1.xml, testMathML.xml, testStructure1.xml, test-382.xml"/>
+        testCorpus1.xml, testMathML.xml, testStructure1.xml, test-382.xml, testPhrase.xml"/>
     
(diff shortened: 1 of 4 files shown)


joeytakeda:
I have now merged the TestFile and build information from #508 and did some slight reorganization, so now this PR is ready for review :-)

HelenaSabel:
The Stylesheets Group considers that the `@` sign should be in the text so one can search and copy the whole string.

sydb (approved):
Turns out the ‘@’ that this PR has rendered by CSS were already being rendered by CSS before this PR. Thus, while I don’t think things are perfect even with this PR, they are certainly better than they were before; furthermore PR seems to work and all tests pass. Thus I think this is ready to be merged.

PR #622, opened 2023-09-12
Only check untyped descs or gloss for out of date translations

raffazizzi:
In support of https://github.com/TEIC/TEI/pull/2473

### DIFF
--- a/common/functions.xsl
+++ b/common/functions.xsl
@@ -1237,8 +1237,8 @@ of this software, even if advised of the possibility of such damage.
     <xsl:param name="context"/>
     <xsl:for-each select="$context">
       <xsl:variable name="lang" select="tei:generateDocumentationLang(.)[1]"/>
-      <xsl:sequence select="tei:desc[@xml:lang='en']/@versionDate gt tei:desc[@xml:lang=$lang]/@versionDate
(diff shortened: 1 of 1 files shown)


sydb (approved):
Code looks reasonable; passes all tests; TEI build works in my docker.

joeytakeda:
I think this has now been handled in https://github.com/TEIC/Stylesheets/issues/635 (PR: https://github.com/TEIC/Stylesheets/pull/636), more or less equivalently: where this PR adds an additional predicate—`[not(@type)]`—that checks just looks for any untyped desc, the current version of `functions.xsl` now specifies excluding deprecationInfo: `[@xml:lang='en' and  not( @type eq 'deprecationInfo' )]` 

So I think we can just close this without merging unless we'd prefer the more defensive `not(@type)`

raffazizzi:
Yeah @joeytakeda I think we attacked this same problem from different angles. I'm pretty sure this is no longer needed.

sydb:
Ah! Here it is; you found it, @joeytakeda. If I could have found this the other day I would have just merged this PR, rather than creating #635 & #636.
I am undecided on whether `not(@type)` is better than what we have (`not( @type eq 'deprecationInfo')`), but think we should close this for now, either way. (And apparently @raffazizzi agrees, as he closed this while I was typing! :-)

Issue #543, opened 2022-02-07
Multiple file extensions are a bad idea


martindholmes:
This ticket is a fresh take on one aspect of what was originally raised in #201: standard Stylesheet ODD processing creates outputs with multiple extensions (file.odd.dtd, file.odd.relaxng), which is pernicious because, for one thing, on Windows, which hides the actual extension by default, it will appear that file.odd.dtd is actually file.odd.  We believe the .odd component should be dropped, and at the same time, the .relaxng extension should be standardized to .rng.

Note that these extra extensions do not occur in the Oxygen framework, so they're not only bad but also inconsistent.

martindholmes:
The other ticket, relating to the /out/ folder in Oxygen, is here: 

https://github.com/TEIC/oxygen-tei/issues/53

martindholmes:
With @HelenaSabel, @sydb, @npcole: We believe the problem arises at lines 212 and 215 of bin/transformtei: the $from variable contains the string "tei" rather than the extension "odd", so the .odd extension is left in place. We think a good approach to this might be a sed or awk command which strips off any of a set of expected extensions (.xml., .odd, .tei etc.) before adding the output extension.

npcole:
The code to strip and replace the filename is wrong.  In my view we should only explicitly strip extensions we are expecting, using a bash for loop to check for them and call basename. Other views are available.

martindholmes:
Stylesheets group believes that we should create teitorng and oddtorng (which should do the same thing), but leave teitorelaxng, making it instead just generate a warning that people should use oddtorng instead.

sydb:
Mostly done with this (in branch sydb_543_file_exts), but the plan outlined above does run into problems.

- What to do if the output filename is the same as the input filename? (I think the answer is to die().)
- What to do if the output file already exists? (I think the answer is to die() unless user has provided --force or --overwrite switch or some such.)

sydb:
Group meeting of 2023-09-28 decides on 2 questions I posed on [2022-09-10](#issuecomment-1242675967):

- die()
- overwrite (And if anyone complains, we can consider adding a `--overwrite` or whatever)

sydb:
This can be closed, EXCEPT that I just discovered a) a dependency in the TEI repo, and in trying to fix that, b) Stylesheets/profiles/tei/relaxng/ should now be called profiles/tei/rng/. Fixed, I hope. (@raffazizzi is testing.)

hcayless:
[...]

PR #628, opened 2023-09-29
delete file extension doubling

sydb:
Per #543, strip off the input file’s extension (if it is one of the extensions we know about, that is) before appending the output file extension.

Code seems to work, but some serious testing is in order. Reviewers should test the code by trying several cases of things like the following, in with no output specified, with an output filename specified that is in the same dir as input, in current dir, or in a different dir altogether.

- Input file in current directory, filetype matches command name
- Input file in different directory, filetype matches command name
- Input file in current directory, filetype a normal one
- Input file in different directory, filetype a normal one
- Input file in current directory, filetype a weird one (or none)
- Input file in different directory, filetype a weird one (or none)

E.g., `cd ~ ; /path/to/Stylesheets/bin/csvtotei ~/rockyracoon.coseva /tmp/rockyracoon.tei` would be an example of the last with output in a different directory altogether.


sydb:
Note — not passing tests yet. (I forgot all about them.) Reviewers should feel free to test now or wait until tests are passing.

sydb:
Craziness — I just discovered that running `cd Stylesheets/ ; make test` _does_ run the clean: target in the Test/Makefile, but running `cd Stylesheets/Test/ ; make` does _not_. 

In any case, I think I _finally_ have the tests working. If anyone is interested, here is how you can rename every file named “A.xml.B.ext” to just “A.B.ext”:
~~~sh
$ for f in *.xml.*; do echo "---------$f:" ; git mv $f ${f%.xml.*}.${f##*.xml.} ; done
~~~

raffazizzi (approved):

Issue #319, opened 2018-04-27
oddtoodd stylesheet generates invalid ODD source


lb42:
As noted in #212 the current oddtoodd process generates invalid ODDs. Specifically:

-  when all attributes have been removed from an attList, the stylesheet generates an empty &lt;attList> rather than nothing
- when a &lt;constraintSpec is added to an attribute, the stylesheets adds it to the compiled spec in the wrong place (ahead of the &lt;datatype> instead of after it)
- a &lt;p> inside a &lt;specGrp> is invalid

Compiling the ODD for teisimplePrint,odd exhibits all these problems; there may of course be others.

martindholmes:
Re the second point: I've never understood why the order of children in an elementSpec is constrained. Is there a reason?

martindholmes:
Eh?

lb42:
Constraining the order of components in an element we are defining doesn't seem such a problem to me. The order is arbitrary, obviously, but requiring that it always be the same makes validation an awful lot simpler. Especially in the absence of features like the ampersand connector, or interleave.

lb42:
see also #210  for other ways of getting an invalid schema out of an apparently correct ODD

lb42:
This issue is now 5yrs old

bwbohl:
Today I also stumbled over the invalid empty `attList` generated when I was compiling my ODD with `odds/odd2odd.xsl`. It's inline element statements that simply create the `attList` elements, no matter what the `$ORIGINAL` or `elementSpec` in the customisation state, e.g.:

https://github.com/TEIC/Stylesheets/blob/2b130e0b1faa400ad648c5bc7c0b9a6a32b391ae/odds/odd2odd.xsl#L972-L978

For a simple local fix I just wrapped it in:

```xml
<xsl:if test="tei:attList or $ORIGINAL/tei:attList">
…
</xs:if>
```

And _voilà_  :-)

sydb:
Thank you, @bwbohl. I think that fix looks right, but (as you point out on the [PR](https://github.com/TEIC/Stylesheets/pull/630)) this should also happen for an `<attList>` in a `<classSpec>`. (And the tests are failing, although I suspect that has nothing to do with your code.) I am pessimistic about getting this into the upcoming release, but have high hopes we can get it into the one after.

lb42:
It has been waiting 5 years...

HelenaSabel:
I took the liberty of assigning this ticket to @trishaoconnor. 
[...]

PR #630, opened 2023-10-04
Fix: creation of empty `attList` in  odd2odd.xsl

bwbohl:
addresses #319

This is a local fix for `elementSpec` and probably these changes should be made in a more general way.

### DIFF
--- a/odds/odd2odd.xsl
+++ b/odds/odd2odd.xsl
@@ -970,15 +970,14 @@ of this software, even if advised of the possibility of such damage.
           </xsl:call-template>
           <!-- attList -->
           <xsl:if test="tei:attList or $ORIGINAL/tei:attList">
-              <attList xmlns="http://www.tei-c.org/ns/1.0">
-                <xsl:apply-templates mode="justcopy" select="tei:attList/@org"/>
-                <xsl:call-template name="odd2odd-processAttributes">
-                  <xsl:with-param name="ORIGINAL" select="$ORIGINAL"/>
-                  <xsl:with-param name="objectName" select="$elementName"/>
-                </xsl:call-template>
-              </attList>
+            <attList xmlns="http://www.tei-c.org/ns/1.0">
+              <xsl:apply-templates mode="justcopy" select="tei:attList/@org"/>
+              <xsl:call-template name="odd2odd-processAttributes">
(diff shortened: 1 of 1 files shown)


sydb:
OK. I have now reproduced the basic change you made for elements (putting the output `<attList>` in a conditional) to attribute lists in classes. I have checked that change into your patch-1 branch (hope you don’t mind), and also merged in the latest from our dev branch while I was there. It is passing all tests (from both Test/ and Test2/) that run on my system.

Can either @bwbohl or @lb42 provide a sample ODD that fails due to an `<attList>` that has had all its attrs removed? (Then we can test it against the new code, here.)

(BTW, still  moderately pessimistic about inclusion in next release, for although this is not a complex fix, there are a dozen others also trying to get into next release :-)

lb42:
The teiSimplePrint odd exhibits this and two other problems according to the issue i originally raised some time ago (#319 ... 5 years ago)

sydb:
OK. I tested using Exemplars/tei_simplePrint.odd (Sorry, @lb42, I had misunderstood). I compared the output of the Stylesheets in this branch against the output of the Stylesheets in the issue_2382_eventName branch (the branch I happened to be on at the time), and the only differences are a) the value of the `@source` of `<schemaSpec>` and b) 86 missing empty `<attList>` elements. 

This leads me to believe this is entirely successful, and could be merged.

An incidental finding is that tei_simplePrint.odd is itself invalid. There are duplicate occurrences of 4 different `@xml:id` values ("page1", "rend-it", "LCSH", & "OTASH").

sydb (approved):

sydb (approved):
Both of the test failures are the SLACK_WEBHOOK thing, so have nothing to do with the actual XSLT. Thus I think if @joeytakeda (or anyone else) can reproduce a successful of an ODD with a known problem, this can just be merged.

lb42:
Those duplicate xmlid values are however all inside egXML elements where arguably they should be permissible.

HelenaSabel (approved):

HelenaSabel:
I made a test with and ODD that had the empty `<attList>` issue and it works great. Thank you @bwbohl!
  
I also ran Test and Test2 in my local machine and everything passed. 

I think this is ready to be merged.

joeytakeda (approved):

Issue #627, opened 2023-09-29
att.repeatable attributes have no effect on `<sequence>` or `<alternate>`


sydb:
I have pretty much convinced myself that `@minOccurs` and `@maxOccurs` do nothing when they are attributes of `<sequence>` or `<alternate>` (as opposed to on `<elementRef>` or `<classRef>`, where they work fine; or on `<anyElement>` or `<datatype>` where I have not performed thorough tests).

I will attach a demo ODD that demonstrates the problem shortly.

Anyone who wants to help with this, please feel free to assign yourself.

sydb:
[issue_627.zip](https://github.com/TEIC/Stylesheets/files/12767655/issue_627.zip)
Why does Github support .zip, but not .tgz? Harumph!

sydb:
Wow. Just noticed that this bug has at least one real sequela in the _Guidelines_:
If you look at the content model for [`<choice>`](https://www.tei-c.org/Vault/P5/4.6.0/doc/tei-p5-doc/en/html/ref-choice.html), you will see that it is defined as **2+** occurrences of either a member of model.choicePart or another `<choice>`:
~~~xml
  <alternate minOccurs="2" maxOccurs="unbounded">
    <classRef key="model.choicePart"/>
    <elementRef key="choice"/>
  </alternate>
~~~
HOWEVER, if you look at the schema declaration, you will see that it is defined there as only **1+** occurrences of either a member of model.choicePart or another `<choice>`:
~~~
element choice {
   ( model.choicePart | choice )+
}
~~~
That has gone unnoticed for many years, I am sure. (The output _should_ be something like either `( ( model.choicePart | choice ), ( model.choicePart | choice )+ )` **or** `( ( model.choicePart | choice ), ( model.choicePart | choice ), ( model.choicePart | choice )* )`, but my code on branch sydb_627 is not quite working, yet.)

PR #632, opened 2023-10-07
Separate pattern for every instance of <anyElement> — fixes #631 (I hope)

### LINKED ISSUE #631: multiple `<anyElement>` results in invalid RELAX NG schema
Another bug that has probably been around for years if not decades and never been noticed. If you have more than 1 `<anyElement>` in a content model the resulting RELAX NG is invalid.

I will post an ODD file (bundled with the output of `teitorelaxng` (soon to be `teitorng`) run on that ODD, as Github makes me ZIP it, anyway) that demonstrates the problem on this ticket shortly.

I _think_ the problem is entirely contained in odd2relax.xsl. It generates an `<rng:define>` for each `<anyElement>` that occurs in the compiled ODD input. Then, in the cleanup phase (in "pass3"), I think it tries to delete some of the `<rng:define>`s that have just been created — presumably all but one. But in this case it deletes them all.

This is some very convoluted, uncommented code. I am still trying to wrap my head around the "max( ancestor::rng:div )" part. An `<rng:div>` has no effect on validation, so why is it important to delete only the `<rng:define>` that is most deeply nested in `<rng:div>`s, and how does that ensure there is only 1 resulting `<rng:define>`?

In any case, this problem is sort of blocking progress on #627, not because it is interfering with writing the _code_ for fixing #627, but because it makes it a pain in the neck to _test_ said code.

Several possible solutions jump to mind, but until I understand at least _what_ the current Stylesheets are doing, better still _why_, I doubt I could really endorse any of them, let alone one over the others:

1. alter "pass3" to delete _all_ `<rng:define>`s for the given pattern, then put one back
2. change code so that only 1 `<rng:define>` per `<anyElement>`-inside-given-`<elementSpec>`-or-`<macroSpec>` is generated in the first place. (This should not be hard to do — change `//tei:anyElement` to `//*[tei:anyElement]` in both places, and change template "anyElement" to match. But who knows what else that might break? Might be a problem because different occurrences of `<anyElement>` _in the same content model_ might have different `@require` or `@except`, no?) 
3. change code so that a separate `<rng:define>` _with a different pattern name_ is generated for each and every occurrence of `<anyElement>`. (This should not be hard to do — change definition of $id in "anyElement" template from `concat('anyElement-',$spec/@ident)` to `concat('anyElement-', generate-id() )`. But changing the _references_ to each such pattern might be hard; probably not, but I am nto sure. And in any case, who knows what else that might break?)

OK, having written that out long hand I take back my previous statement about not knowing which is best. Seems to me at the moment that (3) is probably the only viable solution. Otherwise how would different `@require` and `@except` values in the same content model work?

BTW, I may well continue work on #627 (since it is causing problems for the WWP) first, even with convoluted testing procedures, before paying more attention to this. Thus I have not assigned this issue to a milestone.

sydb:
Generate a different RELAX NG pattern (and reference to it) for every occurence of an anyElement element, and update tests to match. (Previously we were trying to generate one for every content model in which an anyElement occurred, but were generation one for every content model in which an anyElement element occurred only once.)

This is high priority because it is making hard to work on #627.


odds/teiodds.xsl
```diff
@@ -190,6 +190,13 @@ of this software, even if advised of the possibility of such damage.
     </xsl:choose>
   </xsl:variable>
 
+  <xsl:variable name="anyElement_disambiguator_format" as="xs:string">
+    <!-- A string used as the xsl:number/@format to append to pattern names for <anyElement>s (thus to disambiguate them) -->
+    <xsl:variable name="zeroes" as="xs:string+">
+      <xsl:for-each select="1 to ( count( //tei:anyElement ) => xs:string() => string-length() )">0</xsl:for-each>
```
joeytakeda, line 196: I think you may be able to simplify this a bit and just use `format-integer` and set the count as the `$picture` to format `0`: 

```xslt
<xsl:variable name="zeroes" as="xs:string"
 select="format-integer(0, xs:string( count( //tei:anyElement ) ) )"/>
```

odds/odd2relax.xsl
```diff
@@ -220,9 +220,26 @@ of this software, even if advised of the possibility of such damage.
...
-    <xsl:for-each select="$pass2/root">
-      <xsl:apply-templates mode="pass3"/>
-    </xsl:for-each>
+    <xsl:variable name="pass3">
+      <xsl:for-each select="$pass2/root">
+        <xsl:apply-templates mode="pass3"/>
+      </xsl:for-each>
+    </xsl:variable>
+    <xsl:sequence select="$pass3"/>
+    <!--
```
joeytakeda: Happy for this debugging stuff to be left in, but just flagging it in case @sydb didn't mean for it to be retained

joeytakeda (commented):
This looks good to me — one small editorial suggestion for the disambiguator, but otherwise looks good to me (and passes both test1 and 2 on my machine)

odds/teiodds.xsl
sydb: Good point. But format-integer() is an XPath 3 function, and this is an XSLT 2 program. I will not be at all surprised if Saxon will process it correctly even in this XSLT 2 context. But I am not sure if we should rely in that.
On the other hand, why are we still using XSLT 2 programs? Any reason not to just change them to XSLT 3? (What, if anything, would break?)

odds/odd2relax.xsl
sydb: I went back and forth on this one. On the one hand, I generally don’t like leaving commented code in version-controlled documents. Yes, it was useful to have, but now as a comment it just gets in the way. On the other hand, it reminds future maintainers (including me) that this portion of code is called from within an `<xsl:variable>`, and thus you cannot use `<xsl:result-document>`. But, of course, if you ever did uncomment it and use this code, the output RELAX NG would be borked. (Because it would have 4 definitions of most things.)
Writing this, though, I think the right thing to do is to note this _in the program_. I am hoping to check-in an update shortly …

raffazizzi:
Despite the GitHub build error at `Test / test2 (pull_request)`, it all builds correctly on my local machine.

sydb:
OK, can we merge this puppy, then?

HelenaSabel (approved):

HelenaSabel:
In my local machine, I was getting the same ID-generated error, so I updated the expected results.
I think it’s ready to be merged!

Issue #610, opened 2023-05-15
Default conversion docx to tei broken by redundant xml:id 


lb42:
By default, running docxtotei on a bunch of docx files now inserts into the header for each one the following lines:
~~~~
<appInfo>
    <application xml:id="docxtotei" ident="TEI_fromDOCX" version="2.15.0">
     <label>DOCX to TEI</label>
    </application>
   </appInfo>
~~~~
I am not sure what the version number is for though it doesn't seem to be that of the stylesheets package. More seriously the presence of an @xml:id here means that I cannot process a bunch of converted files all together, since the @xml:id values won't be unique. What is the point of this @xml:id anyway? It simply repeats information already given by the @ident attribute.

peterstadler:
I think the `xml:id="docxtotei"` is an anchor for the following error code: https://github.com/TEIC/Stylesheets/blob/a394b17a5951f0c24f1fbe547df9ab61d60f1951/docx/from/functions.xsl#L212-L216

NB, I believe the `@resp` is misspelled here as "teitodocx" – but the rationale seems to output conversion errors in the TEI file and point at the responsible agent.

peterstadler:
I was also looking at the Version issue: I think we could propagate the version info from the shell script https://github.com/TEIC/Stylesheets/blob/a394b17a5951f0c24f1fbe547df9ab61d60f1951/bin/transformtei#L108
to the ANT file and further to docxtotei.xsl. Yet, I'm unsure how to set a proper fallback for users running ant (or the stylesheet) directly?

sydb:
Stylesheets/docx/from/docxtotei.xsl line ~620 needs to have `"2.15.0"` changed to something that reads the Stylesheets/VERSION file.
Stylesheets/docx/from/functions.xsl line ~213 needs to have `"#docxtotei"` (or whatever is right thing) instead of `"#teitodocx"`.

PR #634, opened 2023-10-10
fix errors in docxtotei noticed by Stylesheets group

sydb:
While discussing #610 the Stylesheets group noticed two minor bugs, which this PR is intended to address.

1. Read version from file VERSION (rather than a static hard-coded value that has not been changed since 2.15.0!).
2. Fix typo — function had "#teitodocx" where it meant "#docxtotei" (alright, that is more than a typo, but you get the idea).

### DIFF
--- a/Test/expected-results/hyperlinktest.xml
+++ b/Test/expected-results/hyperlinktest.xml
@@ -20,7 +20,7 @@
     </fileDesc>
     <encodingDesc>
       <appInfo>
-        <application xml:id="docxtotei" ident="TEI_fromDOCX" version="2.15.0">
+        <application xml:id="docxtotei" ident="TEI_fromDOCX" version="7.56.0a">
           <label>DOCX to TEI</label>
         </application>
       </appInfo>
--- a/docx/from/docxtotei.xsl
+++ b/docx/from/docxtotei.xsl
@@ -17,7 +17,7 @@
                 xmlns:r="http://schemas.openxmlformats.org/officeDocument/2006/relationships"
                 xmlns:rel="http://schemas.openxmlformats.org/package/2006/relationships"
                 xmlns:tbx="http://www.lisa.org/TBX-Specification.33.0.html"
-		xmlns:html="http://www.w3.org/1999/xhtml"
+                xmlns:html="http://www.w3.org/1999/xhtml"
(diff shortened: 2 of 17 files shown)


Test/expected-results/hyperlinktest.xml
```diff
@@ -20,7 +20,7 @@
     </fileDesc>
     <encodingDesc>
       <appInfo>
-        <application xml:id="docxtotei" ident="TEI_fromDOCX" version="2.15.0">
```
bwbohl, line 23: To avoid such non-updated version statements, wouldn't it be nice to automatically update them on build or on release and, e.g., just fill in a placeholder here?

ebeshero: @bwbohl I think we've addressed this: We've added a rule to read the current version number in the Stylesheets in docxtotei.xsl.

docx/from/docxtotei.xsl
```diff
@@ -215,473 +198,465 @@ of this software, even if advised of the possibility of such damage.
...
-			</application>
-		      </xsl:if>
-	           </xsl:for-each>
-	        </xsl:if>
-      </appInfo>
-    </xsl:template>
+   <xsl:template name="generateAppInfo">
+     <appInfo>
+       <xsl:variable name="version"
+                     select="if ( unparsed-text-available('../../VERSION') )
```
ebeshero, line 590: @bwbohl Here's where we made the change!

bwbohl: That looks good!
Nevertheless, what I was getting at is manually maintaining TEI files in `Test/expected-results/` vs. having a placeholder there for the value of `@verison` and then, when a text is executed, updating it on the fly from the `VERSION`-file; in the sense of "preparing" or "building" the test prior to execution. Doing so would guarantee that the `expected-results` `@version` is always in sync with the current `VERSION` and avoid the test from failing just due to the value of `@version`.

sydb: Right, @bwbohl. That’s what I thought you might have meant. I think you are absolutely right. But I do not think we plan to do _anything_ to improve comparison to files in Test/expected-results/, because I think we are going to stop using Test/ (in favor of using what is now called Test2/). There we will probably need to change the cleanForDiff target to remove that `@version`.

Issue #635, opened 2023-10-16
failing to handle `<desc type="deprecationInfo">` when testing for out-of-date descriptions


sydb:
I could have sworn we fixed this already, but cannot find an issue or PR about it (whether open or closed).

In common/functions.xsl tei:descOrGlossOutOfDate() is defined as
~~~xml
  <xsl:function name="tei:descOrGlossOutOfDate">
    <xsl:param name="context"/>
    <xsl:for-each select="$context">
      <xsl:variable name="lang" select="tei:generateDocumentationLang(.)[1]"/>
      <xsl:sequence select="tei:desc[@xml:lang='en']/@versionDate gt tei:desc[@xml:lang=$lang]/@versionDate
                         or tei:gloss[@xml:lang='en']/@versionDate gt tei:gloss[@xml:lang=$lang]/@versionDate"></xsl:sequence>
    </xsl:for-each>
  </xsl:function>
~~~
In the case of a structure that is being deprecated via `@validUntil`, it will (or at least should) have both a `<desc xml:lang="en">` to describe the thing being described, and a `<desc type="deprecationInfo" xml:lang="en">`. Thus the XPath on the L side of the 1st `gt` in the `<xsl:sequence>` resolves to a sequence of 2 element nodes, and the function dies as `gt` does not handle sequences.

The fix is to add something like a `[not( @type eq 'deprecationInfo') ]` predicate to the both the XPaths that seek `<desc>` elements (on both L and R side of the `gt`.)

PR #636, opened 2023-10-16
do not check langs of deprecationInfo descriptions

sydb:
Per #635, added clause testing for type=deprecationInfo so we don't get 2+ nodes being tested on one side of 'gt'.


raffazizzi (approved):
Code looks good and build is successful

PR #638, opened 2023-10-23
README creation improvements ala TEI ticket 2070

sydb:
Per [TEI #2070](https://github.com/TEIC/TEI/issues/2070) step 1:

- Do not create links for `<gi>` or `<name type="class">` if version number < 1.0
- Presume class names are encoded with `<name type="class">` (not `<ident>`)
- Generate correct link for those (few) version numbers that are only major.minor, not major.minor.patch (by appending ".0")

**WARNING**
This PR _cannot_ be considered in isolation, it must be combined with the [corresponding PR in the TEI repo](https://github.com/TEIC/TEI/pull/2494).


ebeshero:
I will look at this before the  next release!

profiles/readme/html/to.xsl
```diff
@@ -60,36 +61,62 @@ of this software, even if advised of the possibility of such damage.</p>
...
+  <xsl:param name="version_from_filename" select="replace( $filename, '^readme-([0-9aαbβ.-]+)\.xml$','$1')"/>
   <xsl:param name="vault" select="'https://www.tei-c.org/Vault/P5'"/>
   <xsl:param name="docPath" select="'doc/tei-p5-doc/en/html'"/>
+  
+  <!--
+    The version # we grabbed from the filename might not have the patch level ".0" at the end.
+    But the Guidelines we are trying to point to always have it. So here if we find only major-
+    dot-minor (rather than major-dot-minor-dot-patch), we append a ".0".
+  -->
+  <xsl:variable name="version" as="xs:string">
```
sydb: Umm … wouldn’t this be easier to read as
~~~xml
<xsl:variable name="version" as="xs:string" select="replace( $version_from_filename, '^([0-9]+\.[0-9]+)([aαbβ]?)$', '$1.0$2')"/>
~~~
?

sydb: Yes, it would. Improved in d4b77e36.

ebeshero:
Note: This Stylesheets PR is connected to [Guidelines PR #2494](https://github.com/TEIC/TEI/pull/2494): I'm reviewing them together. I think we have to merge this one in Stylesheets first? @sydb

ebeshero:
Okay--it looks like we need to merge in [the PR on the Guidelines repo](https://github.com/TEIC/TEI/pull/2494) first--to make the revised Readme XML files available for the revised processing here on Stylesheets. Is that the idea, @sydb ?

sydb:
I have not thought about it. I do not think it matters which is first, at least not measured in minutes, beyond a possible temporary build failure. But they should both be merged the same day, or both should be rejected. It would probably be annoyingly problematic on release day if one had been merged and the other had not.

sydb:
I merged the latest from Stylesheets dev branch into this branch (sydb_TEI_2070) and at the same time the latest from TEI dev branch into TEI sydb_2070. Both Test/ and Test2/ passed, and the _Guidelines_ built without error or failure.

PR #648, opened 2023-11-18
overhaul tei:descOrGlossOutOfDate() function

sydb:
This PR is to address what was initially issue 2506 in the TEI repo, but is now #647.

This fix passed all of Test/ and Test2/ in the Stylesheets repo and all of Test/ in the TEI repo. Furthermore, I compared the Guidelines-web/ and Exemplars/ generated by the dev Stylesheets and this branch, and there did not seem to be any significant differences.

I do not know if we want to fix this as usual, and thus it will get released in 7.57.0, or if we want to consider this serious enough for  a patch release, 7.56.1 or whatever. (My instinct is the former, but if it turns out users other than @dlschwartz are getting hit by this …)

### DIFF
--- a/common/functions.xsl
+++ b/common/functions.xsl
@@ -1124,11 +1124,11 @@ of this software, even if advised of the possibility of such damage.
       <xsl:sequence select="tei:makeGloss(.,$langs)"/>
       <!-- now the description -->
       <!--
-	  Change, 2020-02-10 in response to #418:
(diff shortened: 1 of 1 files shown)


sydb:
We have discovered that I already fixed this in dev (at e3aac9d), and we like that fix slightly better than this one. So closing without merging.

Issue #639, opened 2023-10-27
upgrade to 3.0


sydb:
We floated the idea of upgrading to XSLT 3.0 in Stylesheets/ at yesterday’s meeting. Some were in favor of doing so incrementally, on an as-needed basis. While I am not _against_ that idea, I am mildly in favor of all-at-once, perhaps in large part because I am really curious about what happens if we change them all at once. So I investigated.
I checked out a fresh copy of the repo (so there is no release/ directory, and no cruft from previous work I may have done locally), and counted.[1] We have
-  355 version 2.0
-  007 version 1.0
-  001 version 3.0

I then issued
~~~bash
$ xmlstarlet ed -P -L -N l="http://www.w3.org/1999/XSL/Transform" --update "/l:*/@version[.='2.0']" -v "3.0" $(find . -name '*.xslt' -o -name '*.xsl')
~~~
to change every 2.0 to 3.0, and re-counted:
- 356 version 3.0
- 007 version 1.0

Yay!
I then went into the docker, and using that Stylesheets/ dir ran `make test` and `make test2P`; all tests passed (_after_ I issued `export LC_ALL=C.UTF-8; export LANG=C.UTF-8` per @joeytakeda’s suggestion on [Slack](https://tei-c.slack.com/archives/C05V00CDM2M/p1696096399905019?thread_ts=1696093327.912949&cid=C05V00CDM2M)).
Then, using the current dev branch of TEI, I built the _Guidelines_ (still in the docker) with `make clean validate html-web test exemplars teiwebsiteguidelines pdf epub` twice — once using default en, once with OTHERLANGS set to "fr". That failed on the epub target for both languages en and fr. That is, it successfully built clean (duh!), validate, html-web, test, exemplars, teiwebsiteguidelines, and pdf before it died on epub. It died with “Key 'ATTCLASSES' has not been defined; SystemID: file:/usr/share/xml/tei/stylesheet/common/common_tagdocs.xsl; Line#: 2794; Column#: 4”, which is the _exact same error_ I get in with the normal (dev branch) Stylesheets/!

My take away from this experiment is that

1. We (read @peterstadler and @hcayless) need to fix the docker $LC_ALL and $LANG environment variables.
2. We need to fix the epub processing.
3. It is quite reasonable to convert all our 2.0 stylesheets to 3.0.[2]
4. We need to check the 7 stylesheets that are 1.0, too.

**Notes**
[...]

PR #649, opened 2023-11-20
All XSLT 2.0 ⇒ 3.0

sydb:
Council expressed a preference for switching from XSLT 2 to 3 one file at a time.
So here I am submitting a PR which may be rejected just on the principle that it is not in line with the majority of Council’s preferences. But in case folks preferred one-at-a-time in the mistaken belief that it was easier (because it is really _so_ easy to change them all at once), they might appreciate this.

As mentioned on [the ticket](https://github.com/TEIC/Stylesheets/issues/639), I could not use the lovely command I mentioned there, because it does XML processing and therefore messes up the whitespace _inside attribute values_, which is sometimes quite nice to have. So instead I used
~~~sh
$ perl -p -i -e 's,version="2.0",version="3.0",;' $(find . -name '*.xsl' -o -name '*.xslt')
~~~
Note that _every_ file that got changed had 1 and only 1 occurrence of “version="2.0"”, _except_ for ./Documentation/paramform.xsl and ./rdf/make-acdc.xsl which each had 2. 
 
This change passes all of Test/ and Test2/ on my system; `make release` works, and I can build the GLs against these Stylesheets without trouble. (Except for the test of, and build of, PDF output, as I do not have `xelatex` on my system.)


raffazizzi:
Thanks for this @sydb, I was definitely on the one-at-a-time camp because I wasn't sure someone would step up to do this work of changing them all at once, so thanks! It's great to see that all tests pass right away. Given recent events with the release and TEI Garage, I wanted to ask @peterstadler if he thinks this change could have repercussions there? I wouldn't think so since TEIGarage already runs some XSLT3 scripts, but want to make sure

peterstadler:
> […] Given recent events with the release and TEI Garage, I wanted to ask @peterstadler if he thinks this change could have repercussions there? I wouldn't think so since TEIGarage already runs some XSLT3 scripts, but want to make sure

Only @anneferger is able to provide proper feedback here :)
But my guess is too that it shouldn't be a problem. My proposal would be to not wait for TEIGarage to test but rather merge this into dev (when it passes your other tests) and the TEIGarage development shall try with the dev version of the Stylesheets. Since the next Stylesheets release is several months in the future that should be enough time for testing.

anneferger:
I agree with @peterstadler, I'm not expecting any issues but am going to resolve them if they arise :)

PR #650, opened 2023-11-20
ignore `@version` of `<application>` when comparing in tests 

sydb:
I could not just delete `@version` (as was my original plan), because in at least 1 place the output is validated between generation and comparison, and it turns out `@version` is required. Harumph. So the "cleanForDiff" target in Test2/ and the cleanup.pl pgm in Test/ now change it to a ridiculous, but valid, value.


Test2/cleanForDiff.xsl
```diff
@@ -30,6 +30,10 @@
                    
                 4. Remove any timestamps that are generated during processing, since
                    these will be different each time a test is run.
+
+		5. Remove the @version attribute of &lt;application> elements that have
```
raffazizzi: Perhaps this should say "Replace with a placeholder the value of the @version attribute on &lt;application> elements that..."

debian-tei-xsl/debian/changelog
raffazizzi, line 1: This doesn't seem related to the PR, right? It may need to stay here, though, unless you want to re-do the PR; I don't think there's an easy way to remove one file from a pushed commit.

raffazizzi (changes requested):

debian-tei-xsl/debian/changelog
sydb: I have no understanding of why this file pops into our change set intermittently. IIRC @hcayless just said to ignore it, so I do.

Test2/cleanForDiff.xsl
sydb: I have pushed a complete re-wording of that list item.

raffazizzi (approved):

Issue #652, opened 2023-12-12
transformtei help omission


sydb:
The command `bin/teitohtml[5] --help` fails to mention the `--odd` switch.
I.e. line 40 of bin/transformtei should read `dtd|html|html5|rnc|rng|xsd)` instead.
But it’s worse than that. The $format in line 39 references a variable that has not yet been set, so line 40 _never_ works. Line 39 should read `case "$to" in` instead.

martindholmes:
This looks fine to me; unless I'm missing something, it's just user feedback/instructions, so no impact on functionality (although I ran all the tests anyway). Go for merge.

PR #653, opened 2023-12-12
test $to, not $format; include HTML in kinds of outputs that might need ODD processing

sydb:
Per ticket #652.


martindholmes:
All good, no impact on tests.

PR #664, opened 2024-02-23
Fix spelling

jbampton:


sydb:
Same as #663: these changes are insufficient, but fine on their own.

PR #672, opened 2024-03-08
Fix spelling

jbampton:


ebeshero:
@jbampton @trishaoconnor Just a quick comment. Sometimes unusual spellings are part of a specialized discourse. I wondered about this use of "extention" rather than the expected "extension", so I quickly searched and found it in use in the Digital Menota project handbook, on this page down in 7.4.1: https://www.menota.org/HB1-0_ch7_corr.xhtml

We might want to check with some medieval paleographers to be sure it's really a misspelling.

jamescummings:
I don't think that 'extention' is a specialised term in this case, just a misspelling. While it is a variation on extension, it is really 16thC.

sydb:
Well, the mss being described with the word “[extention](https://en.wiktionary.org/wiki/extention)” is a 14th century mss, so it _might_ be the case that the transcription is from the 16th century. But reading the passage it certainly does not sound like it is 16th century, sounds 20th century to me. More importantly, since both source occurrences are in **test** files (Test/test27.xml & Test2/testBibliography1.xml), it does not matter _at all_ how the word is spelled. It may as well be spelled “[Torremolinos, Torremolinos, and complainin’ about the food, oh it’s so greasy, isn’t it](https://www.youtube.com/watch?v=Yz2LaJOVAiA)”.

PR #674, opened 2024-03-23
Fix spelling

jbampton:


sydb (approved):
Passes both Test/ and Test2/ on my system. (Except the one test that requires xelatex, because I do not have that installed.)

PR #675, opened 2024-03-27
Reverse order of `<constraintSpec>` and `<datatype>` in output of odd2odd when merging attributes

sydb:
@trishaoconnor and I found (what we think is) the spot in odd2odd.xsl that caused the output constraintSpec element to appear before the output datatype element, and swapped it with the immediately following constructor in the "odd2odd-mergeAttribute" template. This fixed the error in the output of running odd2odd.xsl on tei_simplePrint.odd, and all tests in both Test/ and Test2/ pass (except I do not have xelatex on my system).


ebeshero (approved):
Appears to be an elegant and useful repair. Thanks!

PR #676, opened 2024-03-28
link fix

bansp:


sydb (approved):
New link works, whereas old one did not; passes GitHub tests; thus I think this is fine to merge.

PR #682, opened 2024-05-06
Fixes #657

### LINKED ISSUE #657: styling of new attr lists
We recently changed the listing of attributes on tagdoc pages to use actual `<ul>` and `<li>` elements, and gave them CSS so that they look _much_ better than they used to. And this seems to work fine in the _Guidelines_ HTML output, but in a customization ODD (e.g., generated by `bin/teitohtml --odd`) the new list is missing the colon and space between the class name and the first attribute.

I think this may be because the CSS file used for custom documentation (https://www.tei-c.org/release/xml/tei/stylesheet/tei.css) does not have a rule for "ul.classSpecAttDefs". But I am not sure whether the rules from this repo (in P5/odd.css) or from the Styleseets repo (in css/odd.css), if either, should be used. (Actually, those specific rules may be the same, even though the files are moderately different. But in any case, those rules apply to `ul.attList a.link_odd + ul.classSpecAttDefs::before`, but the custom documentation HTML has class="link_ref" where the Guidelines have class="link_odd", I think.

joeytakeda:
Adds another rule to odd.css to ensure colon and space


ebeshero:
@sydb @martinascholger @HelenaSabel  I think we just lost track of this work that @joeytakeda did in May to improve the output CSS of generated ODDs. I just updated the branch and the tests are passing. (I'm not suggesting we merge it for the imminent release, but we may want to continue working on CSS for custom ODD-generated HTML on this branch, and merge for the next release?)

HelenaSabel:
@ebeshero: I will be partial to merging the current work as it is after the release (if @joeytakeda agrees). In principle, the current generation of ODD documentation will be superseded relatively soon so I wonder if it isn’t better to work on issues with higher priority.

sydb (changes requested):
I did not investigate _why_, but with this PR there is not just a “: ” after the classname before the list of attributes, there is a “ : ”. (That is, space-colon-space instead of just colon-space.)

joeytakeda:
> I did not investigate _why_, but with this PR there is not just a “: ” after the classname before the list of attributes, there is a “ : ”. (That is, space-colon-space instead of just colon-space.)

I'm not sure what you were seeing here, @sydb — in looking at the changes, I only see colon+space not space+colon+space.

sydb:
Well, I cannot at the moment reproduce the problem I was seeing before, thus approving. (Not sure why this is a draft PR — as far as I am concerned it can be merged.)

sydb (approved):

PR #684, opened 2024-05-11
Fix typo in `tei-to-text.xsl`

jbampton:


peterstadler:
Thanks for sending in this correction!

PR #685, opened 2024-05-16
Update mediawiki-to-tei.xsl

gyulakalcso:
correcting typo (appy -> apply) in row 36


peterstadler:
thanks for spotting and sending in this PR!

PR #688, opened 2024-07-03
skip //name, since it is not supported in the OpenEdition schema

rvdb:


sydb (approved):
The code certainly does what it is advertised to do, thus my approval. But two thoughts that may have you thinking twice:

1. Because the attributes of `<name>` are also dropped, important information (like something being in italics) might be lost.
2. I would sooner see the OpenEdition TEI schema support `<name>`, myself. Do they have a feature request system? (Since the schema has not been updated since 2019, and is in XSD format, I do not have high hopes.)

rvdb:
@sydb You're right:

1. In a subsequent commit (https://github.com/TEIC/Stylesheets/pull/688/commits/eb536a101631ccd720edb2308f09a36c620bcd6a) I've added a mechanism that only skips elements if they don't have a `@rendition` attribute. If they do, a `<hi>` element with according rendition is produced.
2. Yes, there is a Github repository for the OpenEdition scheme, with an issue tracker: https://github.com/OpenEdition/tei.openedition/issues

Issue #687, opened 2024-06-24
Duplicate idents generated when ODDs try to change/replace/delete attributes and reference them in an outdated class


ebeshero:
While addressing PR https://github.com/TEIC/Stylesheets/pull/681 we discovered a problem in ODD processing. When a user has an ODD that references attributes that are no longer defined in that class, this generates duplicate definitions.

When modifying an attribute class, the application of `mode="replace"`, `mode="change"`, `mode="delete"` for an attribute not already in the class being modified generates duplicate attDefs, which are pulled from where that attribute already exists somewhere else. 

This is not an issue for attempting to change attributes that were never defined in the first place.

bwbohl:
I encountered the very same problem, thanks for reporting!

ebeshero:
We have isolated more conditions that cause the problem: It arises:
* When we attempt to remove or change an existing attribute from an attribute class that simply does not contain it,
* **But** the attribute does exist in a different class
* **And** that attribute must be redefined on an element spec with `mode='delete'` (or maybe on `mode='change'`) 

We think we've located the problem and corrected it here by adding some conditional processing to prevent copying the attribute definition from its classSpec, and copying only the change to the elementSpec: https://github.com/TEIC/Stylesheets/commit/1d2ea124d270967e55221611d46be12c856b60c7 

@joeytakeda feel free to elaborate here!

joeytakeda:
Just to add: this _is_ an error in the Stylesheets, but not a new one. The recent change of `@calendar` into its own class revealed the problem since `@calendar` was no longer part of `att.datable` BUT was customized in the old attribute class (e.g. `<attDef ident="calendar" mode="delete"/>` in `att.calendar`) AND the attribute was defined in the elementSpec (for deprecation purposes).

A simpler example, though, might be something like:

```xml
<schemaSpec ident="T" prefix="T_" start="space">
       <moduleRef key="tei"/>
       <moduleRef key="transcr" include="space"/>
       <classSpec type="atts" ident="att.dimensions" module="tei">
             <attList>
                 <attDef ident="resp" mode="change"/>
             </attList>
        </classSpec>
</schemaSpec>
```

Here, we're trying to remove the `@resp` attribute from att.dimensions, which does not contain an `@resp` (but `att.global.responsibility` does). As well, the `space` element currently customizes the definition of `@resp` to provide a customized definition. So when this is processed through the current stylesheets, you get something like (trimmed for brevity):

```xml
[...]

PR #690, opened 2024-07-07
fix complex merging of attrs, #687

ebeshero:
Modify odd2odds so that it does not "justcopy" attDef elements that are not defined in the source class. This is a bit complicated, see our explanation in #687 itself.


ebeshero:
Noting that @HelenaSabel , @sydb , and @ebeshero tested this with a few different ODD scenarios before agreeing to merge this as a much-needed bug fix.

Issue #552, opened 2022-05-28
Makefile creates installed transformtei with potentially incorrect path 


sydb:
Per the comment preceding it, the “/usr” in the following line of the Makefile should read “${PREFIX}”.
~~~make
	perl -p -i -e 's+^APPHOME=.*+APPHOME=/usr/share/xml/tei/stylesheet+' ${PREFIX}/bin/transformtei
~~~
This is currently line 136. See #551 for some discussion.

That said, I do not (at the moment) understand what $APPHOME vs $defaultAPPHOME is doing in the transformtei program. While this should probably be fixed either way, if we understood what was going on we might be able to skip this step entirely.

sydb:
BTW, have to think about how to make sure that the “${PREFIX}” is expanded by the shell, not interpreted by Perl.

bleekere:
Changed `/usr` to `$PREFIX` as proposed in the issue cf. 3bf32eb7ac02a275b1122cb5e3f8038f27bcd4c6. The `$APPHOME` vs `$defaultAPPHOME` may require additional discussion.

sydb:
Oops. Looks like we never generated a PR for this and actually merged it with the dev branch.
Re-opening for more thorough testing (we know the code generates the desired transformtei, but does that desired transformtei do the right thing?) and then creating PR.

ebeshero:
Prodding @sydb to wrap this one up--should be easy?

sydb:
No, _certainly_ not during freeze. I am not sure what this “sydb” idiot was talking about back on 2023-04-04, but @bleekere (appropriately, IMHO) made this change to a branch which was never merged. So this change has never been tested, and looking at it, I think it might be wrong. (As alluded to earlier, one has to be very careful about whether `make` or `perl` resolves the `${}` bit.)
So this is a ticket to put on hold for a 2 weeks and play with after _Balisage_ papers have been submitted.
N.B. 3bf32eb7ac02a275b1122cb5e3f8038f27bcd4c6, the commit in which the change was made in the [issue-552](https://github.com/TEIC/Stylesheets/compare/issue-552) branch.

ebeshero:
@sydb I know! But you'll notice I set the next release milestone on it. I'm basically doing a sort of census of tickets that we'd marked for this milestone and making sure we don't lose track of these... So my comment on this ticket was fully intended for the 7.58 milestone!

sydb:
[...]

Issue #551, opened 2022-05-28
otiose comment in makefile


lb42:
I just had occasion to do a "make install" having updated my local copy of the Stylesheets, when I noticed the following wird comment:
~~~~
# Shouldn't the "/usr" in the following line be /usr ? —Syd & Martin, 2020-07-03
perl -p -i -e 's+^APPHOME=.*+APPHOME=/usr/share/xml/tei/stylesheet+' /usr/bin/transformtei
chmod 755 /usr/bin/transformtei
~~~~
triviial, i know, but what on earth is this supposed to be telling me?

sydb:
Ha!
That one is fun. Thanks, @lb42. My first reaction was “Idunno, mate, and I wrote that comment!”.

But if you look at the Makefile (instead of the output thereof), you’ll see
~~~make
	# Shouldn't the "/usr" in the following line be ${PREFIX} ? —Syd & Martin, 2020-07-03
	perl -p -i -e 's+^APPHOME=.*+APPHOME=/usr/share/xml/tei/stylesheet+' ${PREFIX}/bin/transformtei
	chmod 755 ${PREFIX}/bin/transformtei
~~~

So, importantly, @martindholmes and I were not being nuts asking if A should be A. We were asking if A should be B, which is still a valid question. More on that later.

The next question is why is this comment in the output? I think this is because of the (somewhat crazy, IMHO) syntax of Makefiles that say that when a TAB is the first character of a line, said line is handed to the shell for processing. Indeed, these comment lines (the ‘#’ starts a comment in both most all shells and in a Makefile) are preceded by a TAB. Why the shell decides the comment needs to be reflected in the output, idunno. In any case, I have fixed this by changing the initial TAB character to 8 spaces.

Now on the change @martindholmes & I were musing about. I tested my above change (TAB to SPACE×8) with `make PREFIX=/tmp install`, because I did not want to create an actual installation. The resulting /tmp/bin/transformtei file has on line 11  `APPHOME=/usr/share/xml/tei/stylesheet`. That is obviously wrong, as there _is_ no such directory on this system. There _is_, however, a /tmp/share/xml/tei/stylesheet/ hierarchy, so I am pretty convinced this should be changed. Will make a different ticket, though.

PR #691, opened 2024-07-18
fix path in output transform pgm

sydb:
As per #552, @martindholmes and I discovered an error in the Makefile which caused the transformtei program to be created with an incorrect path-correction. Thus if the user was using the default paths no problem would occur, but a custom path (set by the PREFIX variable) would mess things up.

This PR fixes that problem.

See also #551 for an amusing side-effect of the comment that we added to note this problem.


martindholmes (approved):
Looks good to me.

sydb:
Given that tests pass (not surprising, this change is not tested by the tests, as it is part of the installation process) and @martindholmes approves, I am going to merge & close.

Issue #696, opened 2024-09-20
`<constraintDecl>` from extract-isosch.xsl


sydb:
Now that `<constraintDecl>` is on the[ verge of existence](https://github.com/TEIC/TEI/pull/2594), the Stylesheets need to process it. Since the Stylesheets have two entirely different methods for extracting Schematron from ODD, I have made two tickets. This one is for **odds/extract-isosch.xsl**. (#697 is for **odds/teiodds.xsl.)

PR #698, opened 2024-09-20
handle `<constraintDecl>` from extract-isosch.xsl

sydb:
Addresses issue #696.
Process new `<constraintDecl>` element: use its `@queryBinding` attr, copy over its `<sch:ns>` elements into the declared NS section, and all other child elements into a new declarations section.

### DIFF
--- a/odds/extract-isosch.xsl
+++ b/odds/extract-isosch.xsl
@@ -116,7 +116,7 @@ of this software, even if advised of the possibility of such damage.
         is a direct child of <d:pre>&lt;schemaSpec</d:pre>.</d:p>
     </d:desc>
   </d:doc>
-  <xsl:output encoding="utf-8" indent="yes" method="xml"/>
+  <xsl:output encoding="UTF-8" indent="yes" method="xml"/>
   <xsl:param name="verbose" select="'false'"/>
   <xsl:param name="lang" select="'en'"/>
   <d:doc>
@@ -129,7 +129,9 @@ of this software, even if advised of the possibility of such damage.
   <xsl:param name="teix-ns" select="'http://www.tei-c.org/ns/Examples'"/>
   <xsl:variable name="xsl-ns">http://www.w3.org/1999/XSL/Transform</xsl:variable>
   
-
+  <xsl:mode on-no-match="shallow-copy" name="copy"/>
+  <xsl:mode on-no-match="shallow-copy" name="NSdecoration"/>
+  
   <d:doc>
     <d:desc>Note on keys: should not really need the "[not(ancestor::teix:egXML)]"
     predicate on DEPRECATEDs and CONSTRAINTs, as the elements matched (tei:* and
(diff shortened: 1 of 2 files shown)


odds/extract-isosch.xsl
```diff
@@ -274,8 +262,8 @@ of this software, even if advised of the possibility of such damage.
         <!-- if desired, other NSs can be added manually here -->
       </xsl:variable>
       <xsl:variable name="NSs" select="distinct-values( $allNSs )"/>
-      <!-- For each pair (except those that are empty or are the XLS namespace) ... -->
-      <xsl:for-each select="$NSs[ not(. eq '␝')  and  not( contains( ., $xsl-ns ) ) ]">
+      <!-- For each pair (except those that are empty or are the XSL namespace) ... -->
+      <xsl:for-each select="$NSs[ not( . eq '␝'  or  contains( ., $xsl-ns ) ) ]">
```
joeytakeda: The comment here seems slightly inaccurate, since this is saying all of the `$NSs` except the ones that are equal to the group separating character (␝)(https://www.compart.com/en/unicode/U+241D)or that _contain_ the XSL namespace. I would assume this means that empty ones could conceivably make it through?

odds/extract-isosch.xsl
```diff
@@ -232,21 +228,13 @@ of this software, even if advised of the possibility of such damage.
...
-    </xsl:copy>
-  </xsl:template>
-  
   <d:doc>
     <d:desc>Second pass does most the work ...</d:desc>
   </d:doc>
   <xsl:template match="/" mode="schematron-extraction">
     <xsl:param name="decorated" as="element()"/>
-    <schema queryBinding="xslt2">
+    <xsl:variable name="qb" select="( //tei:constraintDec[ @scheme eq 'schematron']/@queryBinding, 'xslt2')[1]"/>
```
joeytakeda: Typo here, I think — missing the final `l` on constraintDecl

joeytakeda (changes requested):

odds/extract-isosch.xsl
sydb: Uhhh … No to the ultimate question, but maybe phrasing and test should be a little more precise, here.
An “empty” one, i.e. (presumably) an element that has been decorated with both an empty `@nsp` and an empty `@nsu` would end up being the single character `␝`, and thus match the first comparison.
The contains() is meant to ensure that we do not generate a `<sch:ns>` for the XSLT namespace here in the implicit section. (I am not sure why we are worried about it — would it hurt anything if it were generated?) But of course it is not testing equality, so in fact if someone (bizarrely) had a namespace URI of “this-is-not-http://www.w3.org/1999/XSL/Transform_NOT”, this test would incorrectly see it as the XSLT namespace and fail to generate the needed `<sch:ns>`. So I think that should be updated to `matches( ., '␝'||$xsl-ns||'$')`.

odds/extract-isosch.xsl
sydb: Fixed, thank you, good catch.

joeytakeda (approved):

tuurma:
Thanks @sydb and @joeytakeda, merging!

PR #709, opened 2024-11-05
passing rng declarations through to a JSON object as string literal

raffazizzi:
This is needed for minimal RNG support in RomaJS


sydb (approved):
While I am not convinced this is the _right_ way to do this, I am less convinced it is _wrong_, and the change works on this end. (And supposedly works on the Romajs end, too, or @raffazizzi would not have done it.) Thus approval.

peterstadler (approved):
Looks good to me, many thanks!

Issue #712, opened 2024-11-09
Small fix in sorting tei:biblStruct


TomazErjavec:
I tried to generate HTML for an edition, but the conversion fails with
```
Type error at char 10 in expression in xsl:sort/@select on line 821 column 65 of html_core.xsl:
  XTTE1020  A sequence of more than one item is not allowed as the @select attribute of
  xsl:sort (<date>, <date>)
```
the error refers to the line
https://github.com/TEIC/Stylesheets/blob/a74859ca1c98a6ce7d26fcdde175da797b3278ff/html/html_core.xsl#L821

It turns out that some of our imprint elements have more than one date, which is allowed by the standard TEI schema, and also makes sense here: the imprint describes a biblio item that was published in parts in several years. To remove the error, I suggest that the biblStructs are sorted by the first date found in imprint, i.e. change the line to read
```
<xsl:sort select="tei:monogr/tei:imprint/tei:date[1]"/>
```
The same thing would have to be done in latex/latex_core.xsl.

PR #715, opened 2024-11-26
Correct sorting of tei:biblStruct when multiple dates available (#712)

HelenaSabel:
Test and Test2 pass, but actual fix is untested (however, due to the simplicity of the change, it might be unnecessary)

### DIFF
--- a/html/html_core.xsl
+++ b/html/html_core.xsl
@@ -818,7 +818,7 @@ of this software, even if advised of the possibility of such damage.
                               tei:*[1]/tei:author/tei:forename
                               ,tei:*[1]/tei:editor/tei:forename
                               ,'')[1]))"/>
-            <xsl:sort select="tei:monogr/tei:imprint/tei:date"/>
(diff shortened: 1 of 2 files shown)


sydb (approved):
I compared the HTML Guidelines built with this branch vs dev, and found no significant changes (not even to BIB, which surprised me).

The other two `<xsl:sort>` commands in html/html_core.xsl and the other one `<xsl:sort>` command in latex/latex_core.xsl¹ use the `(path/to/nodes)[1]` method rather than the `head(path/to/nodes)` method. I would mildly prefer that they all 5 use one or the other. I don’t really much care which, to be honest. (I had to look up what `head()` does, but as far as I can tell, they do the exact same thing.)

**note**
¹ Which begs the question — why is sorting different in HTML vs LaTeX?

sabineseifert (approved):
It looks good to me and I remember we talked about that in a stylesheets meeting. But I still feel insecure concerning such stylesheets issues and should better not be relied upon.. 😬

Issue #705, opened 2024-10-28
Update Stylesheets `make doc` documentation


ebeshero:
The Stylesheets README includes guidance on building Documentation for this repo by directoring the visitor to enter `make doc`, and outputting the built document at `release/xslcommon/doc/tei-xsl/index.html`. The built materials in `xslcommon` include guidance on how work with Stylesheets on conversions and a "Stylebear" tool (a web form that would let you input parameters to change, and generates a customized XSLT stylesheet), but that mechanism no longer works. (I don't know if I've ever seen the "Stylebear" in action.)

I'm opening this ticket to suggest we update / remove / fix this and other outdated documentation of the Stylesheets. We could remove the documentation entirely, but much of the information is helpful so long as we're still maintaining conversion scripts. But it's full of holes, too: Section 5's "Documentation of common conversions and start points" is a list of broken links. 

Should we be guiding visitors to run `make doc`? If so, what documentation should we be providing in that build now? 

Should we add a warning / caution in our README that some of the contents generated by `make doc` are obsolete?

PR #716, opened 2024-11-26
first teeny step in addressing #705 — make Makefile more readable

sydb:
This PR is intended to change how the "doc:" target in the Makefile **looks**, not what it does. This is just to make it easier to read & understand.

To test, besides the usual `make test` and `make test2P`, run the `make doc` cmd in the Stylesheets released or dev branch, and compare that output (release/xslcommon/doc/tei-xsl/) to the output of `make doc` in this branch. The only differences I found were:
- version numbers
- creation dates
- ChangeLog

i.e., no significant change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,13 +83,23 @@ profiles:
 
 doc: oxygendoc linkcss
 	@echo BUILD Compile documentation
+        # Create the release directory in which to put this stuff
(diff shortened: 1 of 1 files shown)


ebeshero (approved):
This is all much more lucid and should help us with next steps in modifying the "Stylebear".

ebeshero:
I went looking for changes in the output (not really expecting any)--and really, this performs exactly the same in my local dev branch as here. Which means we haven't munged anything. :-) I recommend merging.

PR #721, opened 2024-12-12
Added normalize-space to gloss value

### LINKED ISSUE #717: `<gloss>` (and presumably `<desc>`) content should be space-normalized
Per #2619, a trailing (or presumably, leading) space in a `<gloss>` shows up in the output. (Which output?) If so, I submit there is somewhere where we should be calling `normalize-space()` that we are not.

(Note: I have not added a “conversion: [whatever]” label because I do not know which output(s) are problematic.)

GusRiva:
# Dealing with #717 

## What I did:
Found the place where the content of gloss was created and normalized the whitespace

## How I tested:
- Add whitespace at the beginning or ending of a gloss in /TEI/P5/Source/Specs
- Build the guidelines (using this branch of the stylesheets) and see if the whitespaces are normalized

## Example test:
Change  /TEI/P5/Source/Specs/ab.xml (line 5) to:
`<gloss versionDate="2005-01-14" xml:lang="en"> anonymous block </gloss>`

Old output (with dev branch):

> \<ab\> ( anonymous block ) contains any ...

New output from this branch:

> \<ab\> (anonymous block) contains any ...

### DIFF
--- a/Test/expected-results/test.rng
+++ b/Test/expected-results/test.rng
@@ -4413,8 +4413,7 @@ Suggested values include: 1] spoken (spoken); 2] thought (thought); 3] written (
 Suggested values include: 1] deprecationInfo (deprecation information)</a:documentation>
           <choice>
             <value>deprecationInfo</value>
-            <a:documentation xmlns:a="http://relaxng.org/ns/compatibility/annotations/1.0">(deprecation
-          information) This element describes why or how its parent element is being deprecated, typically including recommendations for alternate encoding.</a:documentation>
+            <a:documentation xmlns:a="http://relaxng.org/ns/compatibility/annotations/1.0">(deprecation information) This element describes why or how its parent element is being deprecated, typically including recommendations for alternate encoding.</a:documentation>
             <data type="token">
               <param name="pattern">[^\p{C}\p{Z}]+</param>
             </data>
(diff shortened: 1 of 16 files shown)


GusRiva:
These changes required some modifications in the test files. 

As you can see, now this normalizes the whitespace of gloss in all ODD transformations. I actually think this is a better output in all the places where it happens. 

However, _if_ we just want to change this in the html guidelines and not as a general rule for gloss,  we need to find a different place to do the change. I think this general solution is probably better, but just in case.

sydb (approved):
Looks good to me.

- I examined the 1-line code change, looks good
- I spot-checked a few expected results, seemed fine
- I ran `make test2P` on my machine — all tests pass
- I ran `make test` in a Docker container on my machine — all tests pass

So I am in favor of merging this ASAR.

sydb:
BTW, @GusRiva , in my little brain the important place to effect this change was in the RELAX NG output, where some tools just report whatever whitespace is there, thus sometimes giving ugly results.

trishaoconnor:
I'm happy to merge this with one reviewer's approval, unless you would prefer to review it first @martinascholger?

martinascholger (approved):
Looks good to me. 

- Run `make test`
- `make html-web` with dev and gusr_717

Issue #723, opened 2024-12-22
`@type` of `<ab>` gets dropped


sydb:
Using the current dev branch, I ran `bin/teitohtml` on a document that had an `<ab type="duckling">`. I was not surprised that the `<tei:ab>` was transformed into an `<html:div>`. I was surprised that the `@type` was dropped. It should (along with `@subtype`) be transformed into a `@class` IMHO.

sydb:
See #724.

PR #724, opened 2024-12-29
output `@class` for an `<ab>` that has a `@type`

sydb:
Addressing issue #723.

Added a small conditional block to the template that matches the `<ab>` element that converts `@type` and `@subtype` to `@class`, if they are present. This is not, IMHO, the right way to do this, but it better matches the rest of the Stylesheets.

I have _not_ created new tests to specifically exercise this code. That said, it passes all of the current Test2/ and that portion of Test/ that properly runs on my system.


sabineseifert (approved):
Looks good to me.

html/html_core.xsl
```diff
@@ -82,7 +82,15 @@ of this software, even if advised of the possibility of such damage.
           </xsl:if>
           <xsl:call-template name="makeRendition">
             <xsl:with-param name="default">false</xsl:with-param>
-          </xsl:call-template>
+	  </xsl:call-template>
+	  <xsl:if test="@type">
+	    <xsl:attribute name="class">
+	      <xsl:sequence select="@type"/>
+	      <xsl:if test="@subtype">
+		<xsl:sequence select="'_'||@subtype"/>
```
HelenaSabel:
```suggestion
		<xsl:sequence select="' '||@subtype"/>
```

PR #725, opened 2025-01-09
improved image scaling in PDF

rvdb:
a small fix for better scaling of images in PDF


ebeshero:
Stylesheets Group recognizes this is a helpful change for JTEI and happily merges it!

Issue #697, opened 2024-09-20
`<constraintDecl>` from teiodds.xsl


sydb:
Now that `<constraintDecl>` is on the[ verge of existence](https://github.com/TEIC/TEI/pull/2594), the Stylesheets need to process it. Since the Stylesheets have two entirely different methods for extracting Schematron from ODD, I have made two tickets. This one is for **odds/teiodds.xsl**. (#696 is for **odds/extract-isosch.xsl**.)

sydb:
See #722.

ebeshero:
Completed with https://github.com/TEIC/Stylesheets/pull/727

PR #727, opened 2025-01-18
actually process `<constraintDecl>`

sydb:
[Note: There is no issue married to this PR.]

Wow. I first found a teeny bug in extract-isosch.xsl which occurred in the case of more than one prefix bound to a particular namespace URI. A warning message told the user that only the 1st prefix would be used, but the code tried to use all of them, and thus failed kinda dramatically.

But while fixing & testing this _tiny_ bug fix (only changes 1 line of code), I realized that our entire processing chain pretty much ignored a `<constraintDecl>` that is in the base (or source) ODD.

So I added code to odd2odd.xsl that will (I hope) merge both the `<constraintDecl>`s from the customization ODD with those from the base ODD on a scheme-by-scheme basis, putting the result in the `<encodingDesc>` of the output compiled ODD. I then added code to extract-isosch.xsl to pay attention to `<constraintDecl>`s wherever they may be.

This code passes both Test/ and Test2/, and I was able to build the WWP schema using this branch. (Which says something, because it is a big customization that has a lot of Schematron.)

### DIFF
--- a/odds/odd2odd.xsl
+++ b/odds/odd2odd.xsl
@@ -551,6 +551,7 @@ of this software, even if advised of the possibility of such damage.
             <xsl:apply-templates select="key('odd2odd-CONSTRAINTDECLS-by-SCHEME', ., $top )"/>
 	  </constraintDecl>
 	</xsl:for-each>
+	<xsl:apply-templates select="tei:encodingDesc/node() except tei:constraintDecl" mode="#current"/>
       </tei:encodingDesc>
     </xsl:if>
   </xsl:template>
--- a/odds/odd2odd.xsl
+++ b/odds/odd2odd.xsl
@@ -559,7 +559,7 @@ of this software, even if advised of the possibility of such damage.
 	<!-- … with any attributes the original <encodingDesc> (if any) had … -->
 	<xsl:apply-templates select="tei:encodingDesc/@*" mode="#current"/>
 	<!-- … and any content the original <encodingDesc> (if any) had; -->
-	<xsl:apply-templates select="tei:encodingDesc/node() except tei:constraintDecl" mode="pass0"/>
+	<xsl:apply-templates select="tei:encodingDesc/node() except //tei:constraintDecl" mode="pass0"/>
 	<!-- … then get the list of schemes to which the <constraintDecl>s apply … -->
 	<xsl:variable name="constraintDecl_schemes" select="$constraintDecls/@scheme" as="xs:string*"/>
 	<!-- … and for each such (unique) scheme … -->
--- a/odds/odd2odd.xsl
+++ b/odds/odd2odd.xsl
@@ -142,6 +142,8 @@ of this software, even if advised of the possibility of such damage.
   <xsl:key name="odd2odd-REPLACE"    match="tei:macroSpec[@mode eq 'replace']"   use="tei:uniqueName(.)"/>
   <xsl:key name="odd2odd-REPLACEATT" match="tei:attDef[@mode eq 'replace']" use="concat(../../@ident,'_',@ident)"/>
 
+  <xsl:key name="odd2odd-CONSTRAINTDECLS-by-SCHEME" match="tei:constraintDecl" use="@scheme"/>
+
   <xsl:variable name="DEFAULTSOURCE">
     <xsl:choose>
       <xsl:when test="$defaultSource != ''">
@@ -498,15 +500,6 @@ of this software, even if advised of the possibility of such damage.
     </xsl:choose>
   </xsl:template>
   
-  <xsl:template match="@*|processing-instruction()|text()|comment()" mode="pass0">
-      <xsl:copy/>
-  </xsl:template>  
-  <xsl:template match="*" mode="pass0">
-    <xsl:copy>
-      <xsl:apply-templates select="@*|node()" mode="pass0"/>
-    </xsl:copy>
-  </xsl:template>
-
   <xsl:template match="tei:elementSpec[@mode eq 'change']|tei:classSpec[@mode eq 'change']|tei:macroSpec[@mode eq 'change']|tei:dataSpec[@mode eq 'change']" mode="pass0">    
     <xsl:variable name="CURRENT_ID" select="generate-id(.)"/>
     <xsl:variable name="CHANGED_SPECS" select="key('odd2odd-CHANGE',tei:uniqueName(.))" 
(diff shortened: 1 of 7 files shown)


odds/odd2odd.xsl
```diff
@@ -534,6 +527,43 @@ of this software, even if advised of the possibility of such damage.
...
+      <xsl:apply-templates select="tei:revisionDesc" mode="#current"/>
+    </xsl:copy>
+  </xsl:template>
+
+  <xsl:template name="insert_encodingDesc_if_needed">
+    <xsl:variable name="constraintDecls" as="element(tei:constraintDecl)*"
+		  select="( document(tei:workOutSource(.))//tei:constraintDecl, //tei:constraintDecl )"/>
+    <xsl:if test="child::tei:encodingDesc or count( $constraintDecls ) > 0">
+      <tei:encodingDesc>
+	<xsl:apply-templates select="tei:encodingDesc/@*" mode="#current"/>
```
joeytakeda: It doesn't look like other `tei:encodingDesc` children are being processed here (e.g. `listPrefixDef` etc). I'm presuming that other encodingDesc children should be included here too, so should it be something like:

```xml
<xsl:apply-templates select="tei:encodingDesc/node() except //tei:constraintDecl" mode="#current"/>
```

odds/odd2odd.xsl
```diff
@@ -534,6 +527,43 @@ of this software, even if advised of the possibility of such damage.
...
+	<xsl:variable name="constraintDecl_schemes" select="$constraintDecls/@scheme" as="xs:string*"/>
+	<xsl:for-each select="distinct-values( $constraintDecl_schemes )">
+	  <constraintDecl scheme="{.}">
+            <xsl:apply-templates select="key('odd2odd-CONSTRAINTDECLS-by-SCHEME', ., document(tei:workOutSource( $top )) )"/>
```
joeytakeda: Unless I'm missing something, the `key('odd2odd-CONSTRAINTDECLS-by-SCHEME')` returns a `constraintDecl`, so wouldn't this end up duplicating the wrapping `constraintDecl`? E.g. should this be:

```xml
<xsl:apply-templates
 select="key('odd2odd-CONSTRAINTDECLS-by-SCHEME', ., document(tei:workOutSource( $top )) )/node()" 
mode="#current"/>
```

odds/odd2odd.xsl
sydb: Oh wow. Good point. (And fast — only ½ hour.) I am not sure how that got dropped. (But I used 
~~~xml
<xsl:apply-templates select="tei:encodingDesc/node() except tei:constraintDecl" mode="#current"/>
~~~
… does the leading `//` matter there?)

odds/odd2odd.xsl
sydb: Uhhh … maybe. Makes sense, but I would want to actually test it to see for sure, which I probably can’t do for next hour or two.

sydb: Yes, you are right. Perhaps worse, though, it was not returning anything because the call to tei:workOutSource() was incorrect. It could be fixed by using `$top//tei:schemaSpec` as the argument, but that makes me worry about what happens if there are more than 1 `<schemaSpec>` (crash, I expect), and how we would select one. Need to think about this a bit more. (Another possibility, of course, is to change how tei:workOutSource() works.)

sydb: Whoo-hoo! I think I have a solution. Basically I just stopped using keys, and hand-assemble the list of `<constraintDecl>`s that are in force.

However, in thinking this through I think I may have stumbled upon a really bad problem. I kinda suspect that much of odd2odd.xsl is messed up, because it is processing stuff that is _not_ in the desired `<schemaSpec>`, but rather is in some other `<schemaSpec>`. (E.g. the key 'odd2odd-CHANGEATT' is set to _every_ `attDef[@mode eq 'change']`, whether it is in the `<schemaSpec>` being processed or another one.)

That is, I think, a solvable problem. The key() function can be handed the `<schemaSpec>` in question as a 3rd parameter whenever it is used. But there may be more to it than that, and even if there were not, this is not something I want to try to get done in < 48 hrs for the upcoming release.

BTW, a side note: the reason this PR works even though I have added this processing to pass0 is that `<constraintDecl>` is not allowed within `<specGrp>`.

ebeshero: So, at this point what could possibly go wrong if we merge this? (Seriously, kinds of problems should we be looking for?)

sydb: What could _possibly_ go wrong is a long, if not infinite, list. But at least nothing we test for in the tests will go wrong. That’s the good news. The bad news is that `<constraintDecl>` does not occur in any Stylesheets/ tests at all, and only in TEI/ tests as a side-effect of being part of the _Guidelines_. It is included in both the _Guidelines_ and one of the Exemplars in the TEI:sydb_use_constraintDecl_and_fr_CMC branch, and works there (as long as you are building against this, the Stylesheets:sydb_fix_constraintDecl_oversight_01 branch).

I think the thing(s) to test is that which we expect users to do — namely, to use `<constraintDecl scheme="schematron">` (both with and without `@xml:lang`, which should make no difference) in either the `<encodingDesc>` or as a child of `<schemaSpec>` (or both, which should generate an error) in their customization ODDs.

sydb: BTW, I think the easy way to test this sort of thing is generate a customization ODD that does what you want to test, and then (using this branch) issue `bin/teitoschematron --odd /path/to/IN.odd /path/to/OUT.sch`. If the OUT.sch looks OK (you may want to compare it to the output you get from same cmd in dev branch), you may want to test that the _rest_ of processing works by using your normal method of generating RELAX NG and HTML, and spot-checking those. (But nothing I did on this branch _should_  change those outputs in any significant way, I don’t think.) Some things to test are whether `@queryBinding` makes it through, that `<sch:rule>`s make it through, and what happens if there are duplicate namespace URIs or prefixes defined in `<sch:ns>`s.

sydb: Note, though, that since #697 has not been addressed, you should not expect to see `<constraintDecl>` output in the RELAX NG output.

sydb:
Well, I have found a bunch of bugs so far …

1. When `<constraintDecl>` is child of `<schemaSpec>`:
    - the section "declarations" in the output has duplicated Schematron
    - the query binding is not carried through
1. When `<constraintDecl>` is a child of `<encodingDesc>`:
   -  the namespaces declared within appear twice each in the "namespaces, declared" section
    - the section "declarations" in the output has TWICE duplicated Schematron
1. When both occur:
    - not really a problem, as this is not allowed
    - our Schematron correctly reports the error
    - that said, it would be nice if XSLT issued an error, rather than generated bogus results

Unless someone suggests a different branch is in order, I am going to start addressing these here on this branch soon. (If anyone feels inclined to step up to the plate and help, it would be greatly appreciated — but your time might be better spent on #697 instead.)

sydb:
The commit I just (inadvertently) pushed is NOT ready for prime-time.

sydb:
OK, this latest version (97822677) is passing both Test/ and Test2/ in Docker. I believe all outright bugs I listed above have been fixed, but it still does not warn user if 2+ `<constraintSpec>`s with same `@scheme` in same file. (It does warn if there is a conflict in `@queryBinding` values, though.)

martindholmes:
ATOP meeting tested today and this is working as intended.

martindholmes (approved):
Good to go.

Issue #737, opened 2025-03-04
Missing double quotes in transformtei cause errors on filenames with spaces


a1ix2:
I'm feeding TEI XML from GROBID which I'm using to process a bunch of pdfs. Those pdf filenames contain space, for example
```
'Zheng et al. - 2015 - Analysis of Hormone–Protein Binding in Solution by.grobid.tei.xml'
```

There are a few places in `bin/transformtei` where double quotes are missing and so it crashes when a filename like the above (and possibly filepaths?). This is my quick fix and I'm not sure it covers all cases or does what it should, but it seems to work right now.

```diff
diff --git a/bin/transformtei b/bin/transformtei
index 0759f5aa..80d354c8 100755
--- a/bin/transformtei
+++ b/bin/transformtei
@@ -160,9 +160,9 @@ else
 fi
 if [ $from = odd ] ; then odd=true; fi
 
-infile=${1}
-indir=$(cd $(dirname ${1}); pwd) # This returns a full path where just $(dirname) might return just ‘.’
-infilename=$(basename ${1})
+infile="${1}"
+indir=$(cd $(dirname "${1}"); pwd) # This returns a full path where just $(dirname) might return just ‘.’
+infilename=$(basename "${1}")
 if [ $# -eq 1 ]; then           # only 1 argument, output file is input with new extension
     # strip off expected suffixes from input filename
     infilebase=${infilename}
@@ -184,16 +184,16 @@ if [ $# -eq 1 ]; then           # only 1 argument, output file is input with new
     # output filename is like input but with common prefix replaced with new one
     outfilename=${infilebase}$outsuffix
 else                            # 2 arguments (if there were more then 2 we would have already aborted)
-    outfile=${2}
+    outfile="${2}"
     # output directory is the directory specified
     outdir=`dirname "${outfile}"`
     # Following line tries to create the output directory if it does not exist yet
     outdir=`(mkdir -p ${outdir}; cd ${outdir}; pwd)`
     # output filename is as specified
-    outfilename=`basename ${outfile}`
+    outfilename=`basename "${outfile}"`
 fi
 
-if [ ${indir}/${infilename} = ${outdir}/${outfilename} ] ; then die "input and output files are the same"; fi
+if [ "${indir}"/"${infilename}" = "${outdir}"/"${outfilename}" ] ; then die "input and output files are the same"; fi
 
 if [ ! -f "${indir}/${infilename}" ] ; then die "Input file $infile does not exist"; fi
```

[...]

PR #738, opened 2025-03-05
fix for filenames with spaces

a1ix2:
transformtei complains when filenames contain spaces:
```bash
~$ src/Stylesheets/bin/teitohtml 'Zhang et al. - 2023 - Chemical Transdifferentiation of Somatic Cells Un.grobid.tei.xml'
basename: extra operand ‘al.’
Try 'basename --help' for more information.
~$ src/Stylesheets/bin/teitohtml Zhang\ et\ al.\ -\ 2023\ -\ Chemical\ Transdifferentiation\ of\ Somatic\ Cells\ Un.grobid.tei.xml
basename: extra operand ‘al.’
Try 'basename --help' for more information.
~$
```

after modifying transformtei as indicated in https://github.com/TEIC/Stylesheets/issues/737#issue-2892661486 everything seems to work:

```bash
~$ dev/Stylesheets/bin/teitohtml Zhang\ et\ al.\ -\ 2023\ -\ Chemical\ Transdifferentiation\ of\ Somatic\ Cells\ Un.grobid.tei.xml
WARNING: No localsource set. Will get a copy from /home/alix/dev/Stylesheets/source/p5subset.xml if necessary.
Convert /home/alix/Zhang et al. - 2023 - Chemical Transdifferentiation of Somatic Cells Un.grobid.tei.xml to /home/alix/Zhang et al. - 2023 - Chemical Transdifferentiation of Somatic Cells Un.grobid.tei.html (tei to html) using profile default lang=en
Picked up _JAVA_OPTIONS: -Dsun.java2d.xrender=false -Dsun.java2d.pmoffscreen=false
     [echo] XSLT generate HTML files (language en, style /home/alix/dev/Stylesheets/profiles/default/html/to.xsl, in /home/alix/Zhang et al. - 2023 - Chemical Transdifferentiation of Somatic Cells Un.grobid.tei.xml, out /home/alix/Zhang et al. - 2023 - Chemical Transdifferentiation of Somatic Cells Un.grobid.tei.html)
Warning: XML resolver not found; external catalogs will be ignored
     [xslt] Found binaryObject without @url.
     [xslt] Found binaryObject without @url.

BUILD SUCCESSFUL
Total time: 2 seconds
~$
```

### DIFF
--- a/bin/transformtei
+++ b/bin/transformtei
@@ -160,9 +160,9 @@ else
 fi
 if [ $from = odd ] ; then odd=true; fi
 
-infile=${1}
-indir=$(cd $(dirname ${1}); pwd) # This returns a full path where just $(dirname) might return just ‘.’
-infilename=$(basename ${1})
+infile="${1}"
+indir=$(cd $(dirname "${1}"); pwd) # This returns a full path where just $(dirname) might return just ‘.’
+infilename=$(basename "${1}")
 if [ $# -eq 1 ]; then           # only 1 argument, output file is input with new extension
     # strip off expected suffixes from input filename
     infilebase=${infilename}
@@ -184,16 +184,16 @@ if [ $# -eq 1 ]; then           # only 1 argument, output file is input with new
     # output filename is like input but with common prefix replaced with new one
     outfilename=${infilebase}$outsuffix
 else                            # 2 arguments (if there were more then 2 we would have already aborted)
-    outfile=${2}
+    outfile="${2}"
     # output directory is the directory specified
     outdir=`dirname "${outfile}"`
     # Following line tries to create the output directory if it does not exist yet
     outdir=`(mkdir -p ${outdir}; cd ${outdir}; pwd)`
     # output filename is as specified
-    outfilename=`basename ${outfile}`
+    outfilename=`basename "${outfile}"`
 fi
 
-if [ ${indir}/${infilename} = ${outdir}/${outfilename} ] ; then die "input and output files are the same"; fi
+if [ "${indir}"/"${infilename}" = "${outdir}"/"${outfilename}" ] ; then die "input and output files are the same"; fi
 
 if [ ! -f "${indir}/${infilename}" ] ; then die "Input file $infile does not exist"; fi
 
--- a/bin/transformtei
+++ b/bin/transformtei
@@ -51,7 +51,7 @@ case "$script" in
         ;;
 esac
 
-echo 
+echo
 echo "  Options, shown with defaults:"
 echo "  --saxonjar=$SAXONJAR  # location of Saxon jar file"
 echo "  --trangjar=$TRANGJAR  # location of trang jar file"
@@ -135,7 +135,7 @@ while test $# -gt 0; do
     --viewportheight=*)   viewportheight="-DviewPortHeight=`echo $1 | sed 's/.*=//'`";;
     --nocompress)       nocompress="-DnoCompress=true";;
     --help)         usageMsg; exit 1;;
-     *) if test "$1" = "${1#--}" ; then 
+     *) if test "$1" = "${1#--}" ; then
            break
         else
            echo "WARNING: Unrecognized option '$1' ignored"
@@ -160,9 +160,9 @@ else
 fi
 if [ $from = odd ] ; then odd=true; fi
 
-infile="${1}"
-indir=$(cd $(dirname "${1}"); pwd) # This returns a full path where just $(dirname) might return just ‘.’
-infilename=$(basename "${1}")
+infile="$1"
+indir="$(cd "$(dirname "$1")"; pwd)" # This returns a full path where just $(dirname) might return just ‘.’
+infilename="$(basename "${1}")"
 if [ $# -eq 1 ]; then           # only 1 argument, output file is input with new extension
     # strip off expected suffixes from input filename
     infilebase=${infilename}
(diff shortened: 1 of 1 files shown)


martindholmes:
Both Test and Test2 are passing. Need to do some more specific tests...

martindholmes:
@a1ix2 I'm testing the actual bin/ symlinks, and finding that the problem with spaces is not solved. For instance, if I do this (in the Stylesheets folder):

`bin/docxtotei tempdir/test.docx`

then everything works; but if I rename the directory:

`tempdir with spaces`

then both of these fail:

`bin/docxtotei tempdir\ with\ spaces/test.docx`
`bin/docxtotei "tempdir with spaces/text.docx"`

Since this is the common use-case, I'm wondering what your use-case was, where the problem was solved?

a1ix2:
@martindholmes I've added a few modifications that should make it work when both filenames and directories contain spaces. I'm extremely rusty on my bash, so I'm following recommendations I found here https://unix.stackexchange.com/questions/118433/quoting-within-command-substitution-in-bash

My use-case I'm trying to convert a ton of TEI XML obtained from GROBID (an academic PDF parser) into a format that is simpler to parse during their ingestion into a RAG. I was looking at `teitohtml` and `teitolite` and was trying the exact commands I included in the first post above.

martindholmes:
Test2 passes, but Test fails for me:

`````
ava -Djava.awt.headless=true -jar ../lib/lib/jing-20120724.0.0.jar -c actual-results/oddbyexample.rnc bare/test2.xml
java -Djava.awt.headless=true -jar ../lib/lib/jing-20120724.0.0.jar actual-results/oddbyexample.xsd bare/test2.xml
fatal: file not found: /home/mholmes/temp/Stylesheets/Test/actual-results/oddbyexample.xsd (No such file or directory)
make[1]: *** [Makefile:590: test-oddity] Error 1
`````

So it seems that oddbyexample.xsd is not being generated. When I run the relevant line manually:

`````
../bin/teitoxsd  --localsource=../source/p5subset.xml actual-results/oddbyexample.odd  actual-results/oddbyexample.xsd
`````

I get files called oddbyexample.xsd.tmp and oddbyexample.xsd.rng, but I don't get a plain xsd file.

This is hard to figure out. It rather looks as though the failure comes when Stylesheets/xsd/postprocess.xsl is being run, but I can't yet figure out what the problem is.

martindholmes:
Meanwhile, both of these still fail:

`````
bin/docxtotei tempdir\ with\ spaces/test.docx
bin/docxtotei "tempdir with spaces/text.docx"
`````

So I don't really know where to go from here. My reading of the situation is that just changing the transformtei script is nowhere near sufficient; we will probably have to work through every stage of every transformation to ensure that paths are being passed along in a safe manner, which will be no small task. @a1ix2 I don't know if you want to take this on, or whether your life would be much simpler if you just batch-renamed all your input files to replace spaces and other unwise characters with underscores.

a1ix2:
Test fails in the dev branch for me, so it's likely unrelated. Test2 fails in the dev branch as well for reasons I can't parse (I'm very new to this).

These two work for me in the a1ix2:double-quotes branch
> ```
> ~/dev/Stylesheets/bin/docxtotei temp\ dir\ with\ spaces/test.docx
> ~/dev/Stylesheets/bin/docxtotei "temp dir with spaces/test.docx"
> ```

sydb:
@a1ix2 and @martindholmes — what shell(s) are you using?

martindholmes:
@sydb I'm using bash. Should that make a difference? We will need to test on zsh as well, of course; thanks for reminding me.

a1ix2:
@sydb I'm using bash as well.

sydb:
I am also using bash. I checked out the a1ix2/double-quotes branch, ran Test/ & Test2/, both seemed to pass. I then tested

1. directory name has spaces, input filename does not
2. directory name does not have spaces, input filename has a space
3. both directory name and input filename have at least one space

I tested all 3 of above with all three escaping mechanisms:

- backslash escaping, e.g. `/tmp/test\ dir\ with\ spaces/`
- double quoting, e.g. `"/tmp/test dir with spaces/"`
- single quoting, e.g. `'/tmp/test dir with spaces/'`

and all combinations worked. (Although, of course, when I accidentally used 2 different escaping mechanisms simultaneously (`"/tmp/test\ dir\ with\ spaces/") it failed.)

martindholmes:
@sydb With a fresh clean checkout of the fork, I get this error running make test:

`fatal: file not found: /home/mholmes/tei/test/Stylesheets/Test/actual-results/oddbyexample.xsd (No such file or directory)`

I'm puzzled that you don't. Test2 passes OK.

ebeshero:
@martindholmes @sydb Reviewing tickets we'd been discussing recently, it seemed we didn't decide what to do here. I tested this locally and it passes Test and Test2 for me.

ebeshero:
And it seems that @sydb has already tested this fix, so it seems to improve things even if it doesn't catch every possible odd combination of slashes and space anomalies. I think we should merge this. . .

ebeshero:
I'm testing the transformation on zsh on my machine just to make sure of the latest version...

ebeshero:
...repeated test transformations of the three conditions above in zsh on my Mac:

* directory name has spaces, input filename does not
* directory name does not have spaces, input filename has a space
* both directory name and input filename have at least one space

Each of these works. So, I think with two positive reviews, we should merge this.

PR #745, opened 2025-05-16
fix #734: Update lists

### LINKED ISSUE #734: Converters to TEI still use type attribute for lists
The Stylesheets still use the rather outdated `@type` attribute to specify lists coming from other formats. This should be updated to use `@rend`.

rettinghaus:
This is an attempt to address #734.

So far everything looks more or less as I expected. I'm just not sure what to make of the `figure class="figure numbered"`.

### DIFF
--- a/Documentation/teixsl.xml
+++ b/Documentation/teixsl.xml
@@ -432,7 +432,7 @@ with a table of links on the left or right. </p>
 whether they should start a new window, occupy all of the current
 window, or stay within the frame. These stylesheets implement the
 following rules:
-<list type="ordered"><item>Any <gi>ref</gi> or <gi>ptr</gi> link stays within the frame</item><item>Any link containing <q>://</q> uses the whole browser window</item><item>Any link starting <q>.</q> uses the whole browser window</item><item>If the stylesheet sets no splitting of the document,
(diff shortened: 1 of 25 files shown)


sydb (approved):
I am not sure what to make of the extra "bulleted" in the `figure/@class`, either, but since it is part of the expected results of a test, and those are the results of the test, it does not strike me as a big problem. (Probably will open another low priority issue for this shortly. — done, see https://github.com/TEIC/Stylesheets/issues/764.)
The file source/p5subset.xml should _not_ have been altered (since it is a generated file), but since it will be updated soon (within a month by @HelenaSabel at the latest) this change will get overwritten.
And that said, we might want to look into a different example to replace gi-item-egXML-fz with one that does not use "unordered" incorrectly. (Probably will open another low priority issue for this shortly. — done, see https://github.com/TEIC/TEI/issues/2736.)

sydb:
This is temporarily blocked by TEIC/TEI#2736 — once that has been handled, we should copy the p5subset generated to the source/ here on this branch, at which time (presuming tests pass) this branch could be merged.

trishaoconnor:
PR https://github.com/TEIC/TEI/pull/2820 has been approved and merged. It is a fix for the originally blocking issue https://github.com/TEIC/TEI/issues/2736 which has now been closed as completed.

Removing the 'status blocked' label for 'status go' for this PR to proceed with re-running the tests to see if this now passes tests.

rettinghaus:
Tests seem to pass now.

trishaoconnor:
Thank you @rettinghaus and @sydb for your work on this PR. Everything looks good to me so I'm happily merging!

Issue #694, opened 2024-08-21
link target for deprecation warning refers to a non-existent page 


peterstadler:
The deprecation warning on pages like e.g. https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-name.html
<img width="1222" alt="Bildschirmfoto 2024-08-21 um 18 35 51" src="https://github.com/user-attachments/assets/29d8a6d9-3ebd-401d-85eb-24baf3214742">

targets http://www.tei-c.org/Activities/Council/Working/tcw27.xml which forwards (301) to  http://teic.github.io/TCW/tcw27.html and finally results in a 404.

I found some references to this URL in `common_tagdocs.xsl` which could be fixed, see https://github.com/TEIC/Stylesheets/blob/0aee6b342f731704d75c61b93edeac572d4b8501/common/common_tagdocs.xsl#L2021

but it maybe just some redirecting issue with the web server?

trishaoconnor:
Update with working link: https://teic.github.io/Documentation/TCW/tcw27.html

trishaoconnor:
Spotted typo in the **Deprecation** section.
 

> <item>An announcement is sent to TEI-L that the Technical Council has decided to depreacate elementName and will drop it as early as YYYY-MM-DD. If anyone strongly objects, they should reply to the list so that the Council might reconsider.</item>

Corrected "depreacate" to "deprecate", committed changes straight to dev.

ebeshero:
Noted that this should redirect to: <https://tei-c.org/documentation/tcw27/>

PR #747, opened 2025-05-19
Update target link to TCW 27 for deprecation warnings 

trishaoconnor:
Updated the target links to TCW 27 for deprecation warnings in [Stylesheets/common/common_tagdocs.xsl](https://github.com/TEIC/Stylesheets/blob/0aee6b342f731704d75c61b93edeac572d4b8501/common/common_tagdocs.xsl#L2021) for #694 

However, still need to check if there may be a redirecting issue with the web server as suggested by @peterstadler.


sydb (approved):
I am not really sure the _Guidelines_ HTML should point to a policy document like this, but certainly if they do it should be to the right document!
Heartily approved.

Issue #749, opened 2025-05-20
do not extract instances of non-existant abstract pattern: Schematron extraction


sydb:
Instances of abstract patterns (which are `<sch:pattern>` elements with an `@is-a` attribute) should not be extracted unless there is also a matching abstract pattern (which would be an `<sch:pattern>` with an `@id` that matches the `@is-a`).

This change is needed because we would like to use abstract patterns in our Schematron (and the TEI branch `issue1981bis` does), but if a customization removes the `<constraintSpec>` that holds the abstract pattern, but not all of the instances of that pattern, then the Schematron will have only the instances, and that is illegal.

ebeshero:
completed with #751

Issue #750, opened 2025-05-20
do not extract instances of non-existant abstract pattern: RELAX NG generation


sydb:
Instances of abstract patterns (which are `<sch:pattern>` elements with an `@is-a` attribute) should not be extracted unless there is also a matching abstract pattern (which would be an `<sch:pattern>` with an `@id` that matches the `@is-a`).

This change is needed because we would like to use abstract patterns in our Schematron (and the TEI branch `issue1981bis` does), but if a customization removes the `<constraintSpec>` that holds the abstract pattern, but not all of the instances of that pattern, then the Schematron will have only the instances, and that is illegal.

sydb:
I did this in the same branch as I did the work for extract-isosch, so this issue should be closed the same time #749 is closed, which should happen when #751 is merged.

I think all conversation should now occur on #751, myself.

ebeshero:
Completed with #751

PR #751, opened 2025-05-20
ignore instances of non-existent abstract pattern

sydb:
This PR splits the template for processing `<sch:pattern>`s into two: one for those with `@is-a` and one for those without. Those without just get copied to the output (with an `@id` add iff needed) as usual. Those with an `@is-a` are only copied to the output if there exists an abstract pattern (i.e., an `<sch:pattern abstract="true">`) whose `@id` matches the `@is-a`.

There are currently no abstract patterns in any of the Schematron in P5 in the dev branch. So if you want to test this against an actual abstract pattern, check out the TEI `issue1981bis` branch and build P5 & run the tests there.

Intended to address #749.


sydb:
1. Note that this work was done in _both_ extract-isosch.xsl (for standalone Schematron extraction) _and_ (just now) in teiodds.xsl (for Schematron-embedded-in-RNG) in this one branch. Thus this PR addresses both #749 and #750.
2. I have no idea why I had to update 4 test files¹ to get the tests to work after I merged dev into this branch. (Well, I know precisely why I had to update them — because the value of `img/@style` now starts with a space … but I do not know _why_ the value of that attr changed.)

_____
¹ The list is:
- Test/expected-results/test-complex.html
- Test/expected-results/test-ota.html
- Test/expected-results/test.html
- Test/expected-results/test.html5

martindholmes (approved):
Reviewed this rather quickly to attempt to solve a build break; looks solid to me.

ebeshero:
This passes Stylesheets Test and Test2 locally. Since we merged `1981bis` yesterday, I'm testing this with a build of TEI from `dev` to see what happens with abstract patterns. Since we how have an abstract pattern defined in the spec for `att.declarable`, I think we need to see whether this validates as expected in the schema as built from TEI `dev` branch.

ebeshero:
@sydb @martindholmes I took some time to test this against a fresh build of the tei_all schema from dev now that we've merged in `1981bis`. The abstract pattern in the schema works as advertised, on a nice `<editorialDecl>` with `<correction>` elements inside. Both the assert and report tests work as we expect. So I think all is well here and I'm going to merge. :-)

PR #756, opened 2025-06-11
Fix for issue #755.

### LINKED ISSUE #755: Test2 should pre-process away changes in the rule numbering of Schematron rules before doing diffs
We already make a point of removing various things that change merely as a result of other meaningful changes, and this is one remaining case where serially-numbered Schematron rules generate lots of distracting differences between actual and expected results, making it harder to assess what actually is different.

martindholmes:
This simply removes the id attributes, since they're not actually required or significant for the purposes of testing.

### DIFF
--- a/Test2/cleanForDiff.xsl
+++ b/Test2/cleanForDiff.xsl
@@ -9,6 +9,7 @@
     xmlns:xlink="http://www.w3.org/1999/xlink" 
     xmlns:dcterms="http://purl.org/dc/terms/"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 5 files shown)


sydb (approved):
I personally would have kept the ID, just trimming off the last numeric bit. But it certainly does not matter except as to make navigation easier on a human, and in truth, it is hard to come up with scenarios where that information is useful even for that.

sydb:
I have manually merged dev into this branch, but for various reasons will not be able to test & commit until tonight, my time.

ebeshero:
Reviewed with @trishaoconnor and @sydb during Stylesheets meeting on June 24. Looks really good to us!

PR #760, opened 2025-06-30
misc: fix spelling

jbampton:


Test2/README.md
sydb, line 1: :heavy_check_mark:

profiles/dhoxss/README.txt
sydb, line 1: :heavy_check_mark:

profiles/iso/schema/svg11.rng
sydb, line 1: I am not sure we should be fixing SVG, even in something this minor. (The mis-spellings were in the original [SVG 1.1 DTD](https://www.w3.org/Graphics/SVG/1.1/DTD/) which, presumably, was converted directly to the RNG we use.)

profiles/oepack/epub/jquery-css-transform.js
sydb, line 1: Looks to me like this is an extension to JQuery that we downloaded. The mis-spelling is in what I suspect is [the original](https://github.com/zachstronaut/jquery-css-transform/blob/master/jquery-css-transform.js), so I think it may be unwise to correct it.

sydb (changes requested):

sydb:
Having not heard from @jbampton I took the liberty of reverting (in a new commit) the changes to the files that are not under TEI control, with the aim of accepting the other changes.

jbampton:
Sorry @sydb I have been so busy.  Working on so many projects.

Thanks again

sydb:
No worries!
And yes, I was kinda taken aback to discover you have some 1,282 repositories. Impressive; not surprised you are busy.

PR #761, opened 2025-07-15
- switched output to OpenEdition-1.6.4

rvdb:
- added support for //titleStmt/(editor|author)/idno

### DIFF
--- a/profiles/jtei/openedition/to.xsl
+++ b/profiles/jtei/openedition/to.xsl
@@ -144,7 +144,7 @@
     <xsl:copy>
       <xsl:apply-templates select="@*"/>
       <xsl:attribute name="xsi:schemaLocation">
-        <xsl:text>http://www.tei-c.org/ns/1.0 http://lodel.org/ns/tei/tei.openedition.1.6.2/document.xsd</xsl:text>
+        <xsl:text>http://www.tei-c.org/ns/1.0 http://lodel.org/ns/tei/tei.openedition.1.6.4/document.xsd</xsl:text>
       </xsl:attribute>
       <xsl:apply-templates/>
     </xsl:copy>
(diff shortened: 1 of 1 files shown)


profiles/jtei/openedition/to.xsl
```diff
@@ -162,6 +162,18 @@
...
   </xsl:template>
   
+  <!-- copy <author>|<editor> <idno> children -->
+  <xsl:template match="tei:titleStmt/tei:author/tei:idno|tei:titleStmt/tei:editor/tei:idno">
+    <xsl:copy>
+      <xsl:apply-templates select="@*|node()"/>
+    </xsl:copy>
+  </xsl:template>
+  
+  <!-- skip <affiliation> <idno> children -->
```
sydb, line 172: ### comment-code mismatch

The comment implies the entire `<idno>` element (when a child of `<affiliation>`) will be skipped. But the code only removes the `<idno>` _tags_, leaving the _content_ of the `<idno>` as part of the output (but the attributes, which might be kinda important, are dropped). I do not know which is the desired result.

sydb (changes requested):
We have no tests to check jTEI⇨OpenEdition. But only looking through the code it all looks quite reasonable, save one minor issue that may well be just a vaguely worded comment rather than a real problem.

profiles/jtei/openedition/to.xsl
rvdb: @sydb OpenEdition doesn't allow `//affiliation/idno`, so I'm indeed just copying its text content without the `<idno>` tag. 
Do you have a suggestion for a more precise wording (in which case I should probably revisit other comments that try to describe the same behaviour)?

sydb: In which case I think this could be merged as-is. But no, I do not have a nice turn of phrase to describe this, so when I have had to comment this kind of process I have gone for something quite cumbersome like “drop the tags of the `<idno>` element, but keep its content” or (perhaps slightly better) “use the content of `<idno>` rather than the `<idno>` itself”. There is probably a better way, but I haven’t found it. :smile: 
I do have a (weak) naming convention for such templates:
- "suicide" = a template like this one, removes the _tags_ but processes the _content_, i.e. the element itself is killed, but its descendants (other than attribute nodes) survive: `<template match="app|choice" name="suicide"><apply-templates/></template>`.
- "familicide" = a template that deletes an entire element including all of its descendants: `<template match="teiHeader" name="familicide"/>`.
- "homicide" = kinda in between — a template that kills itself and all descendant elements, but keeps the string value of its content: `<template match="figure/p" name="homicide"><xsl:value-of select="normalize-space(.)"/></template>`.

rvdb: LOL, I can imagine what a crime scene your stylesheets must look like! If all is fine with this PR, please do whatever you see fit (or tell me if I should take some action first).

PR #762, opened 2025-07-18
copy *:xsl children of //tei:constraintDecl[@scheme eq 'schematron'] …

rvdb:
…to RNG


sydb (approved):
Seems like the right thing to do, and passed all tests on my machine.
(Well, all except PDF→XML, which never works for lack of `xelatex`.)

Issue #746, opened 2025-05-19
Comments inside Schematron rules should be removed when building documentation


martindholmes:
This issue:

https://github.com/TEIC/TEI/pull/2534

was complicated by the fact that comments useful to Guidelines maintainers (but probably not to documentation readers) are currently rendered directly into the Schematron output on the spec page. These should be suppressed, making sure we don't at the same time suppress comments inside egXMLs. The different namespaces (eg vs sch) should make this relatively straightforward.

sydb:
Does sort of beg the question of what we (or any other ODD-writer) would do if we _wanted_ a comment to appear in the output Schematron. Use a `<?tei-comment?>` processing instruction?
Or, of course, we could decide the reverse: comments-in-Schematron get placed in the documentation output, but `<?tei-comment?>` PIs get suppressed. I actually kinda like that, myself. (In truth, though, we already use the name "tei" for comment-like processing instructions, so that might be the better name.)
There are currently only 2 comments-in-Schematron that would need to be changed.

sydb:
Since absolutely no one has weighed in on *how* we should do this, and it is slowing down other tickets, I just implemented as originally suggested — PR #766 drops comments from `<constraintSpec scheme="schematron">` weaving.

sydb:
#766 fixed this.

PR #766, opened 2025-08-11
ignore comments when weaving Schematron

sydb:
This PR address #746.

It passed all tests in a docker and generated reasonable output as far as I could tell. (There are only 2 tagdocs in P5 that have comments in Schematron: `<msDesc>` and `att.declarable`; the latter looks fine, the former is acceptable, but has a bit too much vertical whitespace IMHO.)

I do not think it really needs 4 reviews, just picked a bunch of Europeans, kinda at random.


trishaoconnor (approved):
This change looks good to me and I am happy to approve this PR for merging.

HelenaSabel (approved):

Issue #438, opened 2020-06-04
superfluous and non-XML-conformant file Test/testrend.xml


peterstadler:
This file lacks the closing `</encodingDesc>` tag, wasn't touched for six years, and is not used in any test. I think it can happily be removed!

ebeshero:
I think we've forgotten about this poor but simple Stylesheets ticket... Can one of you finish this off quickly @sydb or @HelenaSabel  or @martindholmes (whoever sees it first)?

trishaoconnor:
Closing as completed as the PR [#768](https://github.com/TEIC/Stylesheets/pull/768) passes tests and is merged!

PR #768, opened 2025-08-15
Remove extraneous file 

sydb:
Remove extraneous file noticed by @peterstadler in #438.

Having removed it, both Test/ and Test2/ pass. Furthermore, I issued `egrep -r 'testrend' *` and there were no hits.


martindholmes (approved):
I'm surprised there aren't more such files lurking. Good to go.

Issue #773, opened 2025-09-05
more trouble with expansion of `@minOccurs` and `@maxOccurs`


sydb:
The PureODD content model
~~~xml
<content>
  <elementRef key="A" minOccurs="0" maxOccurs="2"/>
</content>
~~~
gets expanded to
~~~
( A?, ( A, A? )? )
~~~
which is obviously wrong both because it allows up to 3 `<A>` elements and because it is ambiguous. (When a parser hits the first `<A>` in the document, it does not know whether it is matching the 1st or 2nd 'A' in the content model; RELAX NG does not mind, but both DTD and XSD fall over dead.)
It should be expanded to just `( A, A? )?`.

I am reasonably convinced the problem is in odds/teiodds.xsl in the `<xsl:otherwise>` that starts at line 2242. I think the problem is that the `$min eq 0` case generates an optional A _before_ the rest, but I am still puzzling it through. (Actually, I am now beginning to think only the call to "generateDeterministicOptionals" should be made, not the optional A; i.e. the problem is not that the `$min eq 0` generates that A too early, but that it generates it at all.)

sydb:
More to prove it to myself than anything else, I have tested various cases. These Stylesheets, given the following inputs, generate the outputs that are in the comment on the same line. Those that are erroneous are preceded by a euro sign (‘€’, U+20AC, because it looks a bit like an E for error).
~~~xml
 <elementRef key="A" minOccurs="0" maxOccurs="1"        /> <!--   A? -->
 <elementRef key="B" minOccurs="1" maxOccurs="1"        /> <!--   B -->
 <elementRef key="C" minOccurs="1" maxOccurs="2"        /> <!--   ( C, C? ) -->
 <elementRef key="D" minOccurs="0" maxOccurs="2"        /> <!-- € ( D?, ( D, D? )? ) -->
 <elementRef key="F" minOccurs="2" maxOccurs="2"        /> <!--   ( F, F ) -->
 <elementRef key="G" minOccurs="0" maxOccurs="3"        /> <!-- € ( G?, ( G, ( G, G? )? )? ) -->
 <elementRef key="H" minOccurs="1" maxOccurs="3"        /> <!--   ( H, ( H, H? )? ) -->
 <elementRef key="I" minOccurs="2" maxOccurs="3"        /> <!--   ( I, I, I? ) -->
 <elementRef key="J" minOccurs="3" maxOccurs="3"        /> <!--   ( J, J, J ) -->
[...]

Issue #354, opened 2019-01-06
minOccurs and maxOccurs problem


sydb:
I just tested and found that
~~~~~xml
<elementRef key="title" minOccurs="1" maxOccurs="8"/>
~~~~~
seems to work fine, but
~~~~~xml
<elementRef key="publisher" minOccurs="0" maxOccurs="3"/>
~~~~~
incorrectly spits out 
~~~~~
     (t_publisher?,
      (t_publisher, (t_publisher, t_publisher?)?)?),
~~~~~
That is, there are 1 too many t_publishers there. 

See #135, #174, and maybe #297.

Issue #297, opened 2017-11-20
anyElement/@minOccurs and anyElement/@maxOccurs ignored


sydb:
Title kinda says it all. But to be clear, I put
~~~~~xml
  <anyElement require="http://www.example.org/useme"
              minOccurs="1" maxOccurs="unbounded"/>
~~~~~
in an ODD, and what came out was just
~~~~~xml
  <rng:ref name="anyElement-SPECNAME"/>
~~~~~
i.e., the occurence wrapper element `<rng:oneOrMore>` was missing.

sydb:
Note: this is a blocker of ticket https://github.com/TEIC/TEI/issues/1716.

sydb:
The error looks to be in teiodds.xsl ~331, the template that matches "anyElement" in "tangle" mode. It just output a single `<rng:ref>`. I’ve changed it to wrap that `<rng:ref>` in tei:generateIndicators(@minOccurs,@maxOccurs). I will check it in as soon as my local build is successful or my boarding group is called, whichever comes first.  :-)

sydb:
Pushed fix at be2a2e8. I think it worked. However, my commit msg is messed up. Is there a way to edit my commit message after the fact?
This ticket should be closed as soon as Mr. Jenkins is happy and someone tests result.

Issue #174, opened 2016-09-28
Pure ODD minOccurs maxOccurs not respected in RNG


jamescummings:
If I create a new element with a content model of:

```
<content><elementRef key="name" minOccurs="5" maxOccurs="7"/></content>

```

(for example, but any numerical range) then this should mean that I must have 5, 6, or 7 'name' elements. What happens in the RNG is that it requires me to have 5 (fine), but complains at 6 or 7 instances of 'name' elements.

Clearly this is an artificial example.  Am I misunderstanding the attributes, or should this work?

lb42:
This is a known bug.

jamescummings:
Closed. Fixed with #135

PR #774, opened 2025-09-05
fix for `@minOccurs` of 0 and `@maxOccurs` of 2 or more (but not "unbounded")

sydb:
Wow. We have been banging away at this for _years_, and I think we may finally have it right. (If you need convincing, see #135, #174, #354, and maybe #297.) This fix addresses #773.

I think if one or two people review this (at least in part by checking out this branch and trying some ODDs that have various `@minOccurs` and `@maxOccus` on various things) and find no problems, it is good to merge. (I.e., I do not think all 6 reviews need be in before merge.)


raffazizzi:
FWIW I could only reproduce this issue on `elementRef` and `classRef` (to a model class) and this branch **fixes** the problem as far as I can tell.

Works on both dev and this branch:
* `<alternate minOccurs="0" maxOccurs="2">`
* `<datatype minOccurs="0" maxOccurs="2">`
* `<sequence minOccurs="0" maxOccurs="2">`
* `<anyElement minOccurs="0" maxOccurs="2"  except="http://www.tei-c.org/ns/1.0">`

raffazizzi (approved):

trishaoconnor:
Tested using this odd file with @sydb and all looks well.  [test774odd.txt](https://github.com/user-attachments/files/22601819/test774odd.txt)

Looks like it is ready to merge!

Issue #187, opened 2016-11-18
profiles/default/tei-fo.xsl is mysterious and probably irrelevant


martindholmes:
Anything relating to the fo profile should be in profiles/default/fo, where a file called to.xsl does exist. This strange file one level up the tree does the same as profiles/default/fo/to.xsl (which is actually nothing, except to import Stylesheets/fo/fo.xsl), so it looks like a stray file that should not be needed.

martindholmes:
Deleting the file doesn't cause any problems for the Stylesheets build.

sydb:
In a non-comprehensive search do not find any reference to this file from any XSLT, makefile, or ant task in the Stylesheets/ repo.

martindholmes:
@ebeshero and I confirm that the file profiles/default/tei-fo.xsl still exists, and is not linked from anywhere. It should go. Its history suggests it was added in error in 2009.

trishaoconnor:
@ebeshero and @syd have approved PR #775, it has been successfully merged, so I'm closing this ticket as completed.

PR #775, opened 2025-09-15
Deletion of unused file per issue #187.

martindholmes:
This just looks like a simple one that got neglected.


sydb (commented):
Looks good to me. (This file essentially matches /fo/to.xsl, and was probably just a mistake.)

sydb (approved):

Issue #452, opened 2020-08-05
triage profiles


peterstadler:
There are several directories under 'profiles' which seem to be outdated and even might pose a security risk (see #395, #433). This issue is a stub for reviewing all profiles and see what actions can be taken.

npcole:
Review of directories:

* profiles/acm — last updated by Sebastian Rahtz in 2014.
* profiles/adams — is this still used by the Adams Papers Project? Last updated in 2013.
* profiles/agora — Looks like a project run by Lou Bernard, but I can't find information about it online.  Last updated in 2021 by @HelenaSabel — could we have more information about this in the ReadMe file?
* profiles/default — we need this!
* profiles/dhonvalidator — no information in Readme file.  @lb42 can you tell us any more about what this is for?
* profiles/dhoxss — something to do with the Oxford DH Summer School. Last changed 2014 by Sebastian Rahtz.  Is this still in use for anything?
* profiles/enrich — "EU project on manuscript description" — is this still in use?  Last substantive commit by Sebastian Rahtz 2014.
* profiles/iso — "ISO Documentation".  Last substantive changes specific to this project in 2014? Is it still in use? Looks like it was sophisticated.
* profiles/jsi — http://nl.ijs.si  appears to be a live project.  Presumably this profile is still in use.
* profiles/jTEI — in active use.
* profiles/oepack — Last changes specific to this profile in 2014.  
* profiles/ota — in active use? Last changes specific to this profile by SR in 2015.  
* profiles/otapages — last changes specific to this profile in 2014.
* profiles/oucscourses — almost certainly no longer in use.  Last changes 2014. OUCS no longer exists.
* profiles/oulipo — last changes in 2017. Readme says no longer in use.  Remove?
* profiles/oxford — Oxford University IT Services.  Not clear what it is used for. Last changes 2017.
* profiles/readme — Readme format for TEI releases. 
* profiles/saxonce — last updated 2014. Saxonce no longer in Development.  Remove?
* profiles/sciencejournal — experimentation with conversion from word.  Last worked on in 2015 by SR.
* profiles/simple — last updated 2014.  Originally to support TEI Simple project.  Still in use? Looks for a folder that does not exist!
* profiles/tcp — last actively worked on in 2014. For "TCP Texts". Is this still being used by https://textcreationpartnership.org/tcp-texts/ ?
* profiles/tei — stet.
* profiles/teikindle — Last updated 2014.  
[...]

PR #776, opened 2025-09-15
Issue #452: First few profiles removed, and documentation added.

martindholmes:
This is a pilot PR so people can confirm that the approach I am taking is acceptable.


README.md
```diff
@@ -63,3 +63,7 @@ and tool-based projects such as
 [Anthologize](http://anthologize.org/), 
 [Versioning Machine](http://v-machine.org/), 
 and many more diverse projects.
+
+## Deleted profiles and transformations
+
+Some obsolete profiles and transformations have been removed. See DELETED.md for reasons, and details of the last release which included them.
```
sydb, line 69: Should “them” be “each”?

sydb (approved):

README.md
martindholmes: Maybe "each of them". Doesn't seem a big difference though.

I'm no longer sure that Council wants me to do this stuff, so we should run it by them at the next meeting.

martindholmes: Maybe "each of them". Doesn't seem a big difference though.

I'm no longer sure that Council wants me to do this stuff, so we should run it by them at the next meeting.

PR #782, opened 2025-10-09
Fix Illegal character in ODD path on Windows

daliboris:
Fixes issues on Windows machine, where `$currentDirectory` variable contains path in Windows format, for example `C:\Data\Works`.

This type of path occurs, for example, in the oXygen XML Editor, where the path is passed to ANT in the `inputFile` parameter as the editor variable `${cf}`.

In Windows, this led to the following error:

```text
ransformation failed. 
Fatal error during transformation using C:\Users\Boris\AppData\Roaming\com.oxygenxml\extensions\v27.1\frameworks\tei\tei\xml\tei\stylesheet\odds\odd2odd.xsl: Relative URI {C:\Data\Works...} is invalid: Illegal character in opaque part at index 2: C:\Data\Works\odd/sample.odd; SystemID: file:/C:/Users/Boris/AppData/Roaming/com.oxygenxml/extensions/v27.1/frameworks/tei/tei/xml/tei/stylesheet/odds/odd2odd.xsl; Line#: 247; Column#: 25
```

### DIFF
--- a/odds/odd2odd.xsl
+++ b/odds/odd2odd.xsl
@@ -240,6 +240,9 @@ of this software, even if advised of the possibility of such damage.
           <xsl:value-of select="$currentDirectory"/>
           <xsl:value-of select="$loc"/>
         </xsl:when>
+        <xsl:when test="matches($currentDirectory, '^[A-Z]:\\')">
+          <xsl:value-of select="string-join(('file:///', translate($currentDirectory, '\', '/'), $loc), '/')"/>
+        </xsl:when>
         <xsl:when test="$currentDirectory=''">
           <xsl:value-of select="resolve-uri($loc,base-uri($top))"/>
         </xsl:when>
--- a/odds/odd2odd.xsl
+++ b/odds/odd2odd.xsl
@@ -241,7 +241,8 @@ of this software, even if advised of the possibility of such damage.
           <xsl:value-of select="$loc"/>
         </xsl:when>
         <xsl:when test="matches($currentDirectory, '^[A-Z]:\\')">
-          <xsl:value-of select="string-join(('file:///', translate($currentDirectory, '\', '/'), $loc), '/')"/>
+          <xsl:value-of select="string-join(('file://', if(matches($currentDirectory, '^[A-Z]:\\$')) 
+            then  translate($currentDirectory, '\', '') else translate($currentDirectory, '\', '/'), $loc), '/')"/>
         </xsl:when>
         <xsl:when test="$currentDirectory=''">
           <xsl:value-of select="resolve-uri($loc,base-uri($top))"/>


sydb:
Hi @daliboris — good to see you in Kraków!

Hmmm … my oXygen (27.1) does not list `${cf}` when I try to set up a new transformation scenario and click on the little “Insert Editor Variables” button (a yellow ellipsis under a green down-arrow). And both the field names¹ and the descriptions² of the variables I do see suggest that the values here are URLs, not filepaths. Since URLs use ‘/’ not ‘\’, this would seem like it might be a bug in oXygen. But if they do have a variable correctly returning a path (and your error msg sure makes it look that way) I think such a fix is quite reasonable.

Does it work? I do not have any way to test it. I worry that the `string-join()`, because it is sticking a slash in between each member of the sequence, will return a string that starts with `file:////` instead of the desired `file:/` or `file:///`. (I am not sure it would make a difference on a Windows system, though.)


**notes**
¹ “XML URL” and “XSL URL”.
² E.g. “cfdu = The path of the current file directory (URL)” and “CurrentFileURL = The path of the currently edited file (URL)”.

sydb:
I have assigned this to myself, and given it a BUG label, but not a CONFIRMED label because I do not have a Windows system and thus cannot reproduce the error. But I do not plan on merging this until somebody who is not @daliboris reviews and verifies that it works properly on Windows. I haven’t assigned any other reviewers, though, because (in general) I do not know who has easy access to a Windows system to test.

daliboris:
> Hi @daliboris — good to see you in Kraków!

I'm glad I was part of this nice event.

> Hmmm … my oXygen (27.1) does not list `${cf}` when I try to set up a new transformation scenario and click on the little “Insert Editor Variables” button (a yellow ellipsis under a green down-arrow). 

Here are screenshots from my laptop copy of existing transformation scenario without any modifications:

oxygen27 1_2025-10-10T00-40-16

oxygen27 1_2025-10-10T00-40-19

oxygen27 1_2025-10-10T00-40-39


[Here](https://www.oxygenxml.com/doc/versions/27.1/ug-editor/topics/editor-variables.html) is list of all available variables.

> Does it work? I do not have any way to test it. I worry that the `string-join()`, because it is sticking a slash in between each member of the sequence, will return a string that starts with `file:////` instead of the desired `file:/` or `file:///`. (I am not sure it would make a difference on a Windows system, though.)

You are right, it's my fault: the code adds extra slashes, but surprisingly local paths on my PC works also with these "extra" slashes.

This version returns the correct file URI:

```xml
<xsl:value-of select="string-join(('file://', if(matches($currentDirectory, '^[A-Z]:\\$')) 
              then  translate($currentDirectory, '\', '') else translate($currentDirectory, '\', '/'), $loc), '/')"/>
```

> **notes** ¹ “XML URL” and “XSL URL”. ² E.g. “cfdu = The path of the current file directory (URL)” and “CurrentFileURL = The path of the currently edited file (URL)”.

sydb:
Oh! So it is a parameter — and look at that, it is documented as a path, not a URL. I learned something new, thank you.

Do you want to commit the extra-slash-that-is-technically-incorrect-but-seems-to-work fix? If not, I think gitHub will actually let me commit something to your branch, but I am not 100% on that.

daliboris:
Windows path to URI functionality has now been fixed. 

Thank you for your prompt cooperation.

HelenaSabel:
@daliboris: we tried to reproduce the error in two Windows machines, one with Oxygen 27.1 and another with Oxygen 28.0 and we didn’t get any errors... Could you tell us how  to reproduce the error or would you like to participate in our Stylesheets meeting from March.

PR #785, opened 2025-10-22
jTEI: split long urls in pdf

rvdb:
added zero-width space before &amp; in URLs that are displayed in the running text, in order to create break-points for the XSL-FO formatter


martindholmes:
Might be worth doing it before or after forward-slashes too. Lots of long urls don't have any ampersands.

rvdb:
Great suggestion, Martin. I've checked, and following characters seem to be treated as splitting points already by FOP:

- forward slash: `/`
- hyphen: `-`
- question mark: `?`

...that's probably why I hadn't felt the need until I'd encountered this really long URL with only request parameters in a jTEI article. Do you think it would be wiser to add zero-width spaces before them as well?

martindholmes:
Ah, I didn't realize that slashes worked as breakpoints. I've been working with a very old version of Xep, so I'm in the habit of adding those spaces around any path-delimiter because it needs them.

PR #788, opened 2025-11-19
improve Test/ process — optionally move comparisons to end of Makefile itself

sydb:
First crack; seems to work, although output is still a bit ugly.

This PR changes only 1 file, Test/Makefile. The default behavior should be pretty much the same, but the DIFFNOW parameter has been ditched in favor of a new DIFFEND parameter, which defaults to 0.

When set to 0, this new parameter causes the Makefile to follow the current default behavior: every time a comparison is required, a `diff` is executed. The entire `make` fails on the first comparison that finds differences.

When set to 1, this new parameter causes the Makefile to build a temporary bash executable file (/tmp/DIFFEM.bash by default) that contains a sequence of all the `diff` commands that _would_ have been executed during the build, and runs it as the last thing the Makefile does. The `make` does not fail on diff errors at all, but they show up in a somewhat reasonable format clumped together at the end.

I have requested review from the three Jenkins maintainers, as side effects in the CI server is the thing I am most worried about.

While this would formally be low priority, I think it is _so_ much easier, I suspect @trishaoconnor would greatly appreciate it if we could have this merged by early December.

Note: The _internal_ documentation in the Makefile itself has been updated, but the [Update P5 Subset](https://docs.google.com/document/d/1UhCg76_AgnS5PW0g_EN6zPZE4tTUs0bNeMXFVmZ4VuQ/edit?tab=t.0#heading=h.7fm4gojxhcp7) document will also need to be updated if & when this PR is merged. (So please either update said documentation or ping me to do so when this is merged.)

### DIFF
--- a/Test/Makefile
+++ b/Test/Makefile
@@ -12,9 +12,9 @@
 # aftwer we are done with everything else?
 DIFFEND=0			# 0 = do diff now; 1 = do diffs at end of processing
 
-DIFFSWITCHES=" "		# switches user can pass in to # override — not yet implemented
-S4D1=-C0 -bBw			# switches used for most calls to diff
-S4DR=-C0 -bBw -r		# switches used for recursive diff
+DIFFSWITCHES=     		 # switches user can pass in to override
+S4D1=-C0 -bBw $(DIFFSWITCHES)	 # switches used for most calls to diff
+S4DR=-C0 -bBw -r $(DIFFSWITCHES) # switches used for recursive diff
 
 DIFFOUT=./DIFFALL.bash	# path to output program to be run (iff DIFFEND=1)
 
@@ -659,5 +659,11 @@ oddtree:
 dodiff:
 	@echo "--------- Test/Makefile target=$@"
 	@if [ -f $(DIFFOUT) ]; then chmod a+x $(DIFFOUT); fi
-	@if [ $(DIFFEND) -eq 1 ]; then echo " "; echo " "; echo "========= comparisons ========="; fi
-	@if [ $(DIFFEND) -eq 1 ]; then bash -c $(DIFFOUT); fi
+	@if [ $(DIFFEND) -eq 1 ]; then                       \
+		echo " ";                                    \
+		echo " ";                                    \
+		echo "========= comparisons =========";      \
+	        bash -c $(DIFFOUT);                          \
+		echo " ";                                    \
+		echo "If you would like to repeat the comparisons, issue \`$(DIFFOUT)\` at the commandline in this directory."; \
+		fi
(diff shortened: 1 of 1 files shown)


martindholmes:
Currently the build fails if run without the DIFFEND=1 flag because it's trying to read /tmp/DIFFEM.bash, but that file is not created.

@HelenaSabel and I also think the file should be created in the Test folder somewhere, not in /tmp/, which may not be there on any given system.

martindholmes:
Looks good to me now. All working.

Test/Makefile
```diff
@@ -9,9 +9,17 @@
...
-# done building them here)?
-DIFFNOW=1			# 1 = do diff now; 0 = don't do diffs now
+# aftwer we are done with everything else?
+DIFFEND=0			# 0 = do diff now; 1 = do diffs at end of processing
+
+DIFFSWITCHES=" "		# switches user can pass in to # override — not yet implemented
+S4D1=-C0 -bBw			# switches used for most calls to diff
+S4DR=-C0 -bBw -r		# switches used for recursive diff
+
+DIFFOUT=./DIFFALL.bash	# path to output program to be run (iff DIFFEND=1)
```
peterstadler, line 19: I see that you already updated this line, but here's yet another proposal:
Using `mktemp` will actually create the file (not only set its name) and return its filename (absolute path). This will make sure the file exists.
```suggestion
DIFFOUT!=mktemp	# path to output program to be run (iff DIFFEND=1)
```

peterstadler:
another caveat: If you really want to hard code the filename it needs to be added to the `.gitignore` file.

Test/Makefile
sydb: Yeah, I am a bit of a fan of `mktemp` (I say “bit of” because in theory I like it, but in practice I always have to re-read the man page; if I were _really_ a fan, I wouldn’t need to do that : -), but if we use a dynamically generated name, we have no mechanism for having `make clean` remove the temp file. (Or of putting it in .gitignore, as per your next comment.)

But, he said thinking aloud, what if we used
~~~make
DIFFOUT := $(mktemp TestMakefileTempXXX.erasethistempfile)
~~~
we could then have `make clean` perform `rm *.erasethistempfile` and add `*.erasethistempfile` to .gitignore, no?

peterstadler: isn't it enough to do `rm $(DIFFOUT)` for removing the tempfile again?

Also, a dynamically created file by `mktemp` usually is not created in the current working directory so we don't need to exclude it in `.gitignore`.  That said, there's nothing against providing the suffix  (to the `mktemp` command) and excluding this suffix in `.gitignore`

sydb: Well, if we put `rm $(DIFFOUT)` in this target, the user would not be able to easily re-run the comparisons, which could be quite useful. The idea is that `make clean` should remove that file. Thus it has to be remembered across invocations of `make`, which is why I shy away from `mktemp`. (But using `--tmpdir=.` would solve the not-in-this-directory part, but not the name part.)

Which raises the point that it would be a good idea to have this target end by telling user how to do that; something like “if you want to repeat the comparisons only, issue ./$(DIFFOUT) from the Stylesheets/Test/ directory”.

joeytakeda:
With Stylesheets Group, tested locally and works great -- merging

PR #790, opened 2025-11-26
Fix for issue #789.

### LINKED ISSUE #789: Provide a default value for romanFont in latex_param.xsl
Based on this discussion:

https://github.com/TEIC/TEI/issues/2842

I believe we should provide a default value for the romanFont parameter in the latex_params.xsl file, and I would suggest Times New Roman as a good choice, since it covers a decent range of codepoints (including the Greek from that issue), and is widely available. Another option would be Times, since it's one of the 14 core PDF fonts, but according to my reading of the spec (https://opensource.adobe.com/dc-acrobat-sdk-docs/pdfstandards/pdfreference1.3.pdf), it's a type 1 font which covers only the win ansi 1252 range.

Again according to my reading, not providing any value here would cause the processor to use Computer Modern typeface family, designed by Donald Knuth as part of the original TeX typesetting system; this also presumably has a relatively small range.

martindholmes:
This provides what I think is a useful default value for the romanFont parameter. @sydb Note that there were no errors in Test2, which suggests an area lacking coverage (specifically, Test2 is not doing diff checks of any generated .tex files).

### DIFF
--- a/Test/expected-results/test.tex
+++ b/Test/expected-results/test.tex
@@ -17,6 +17,8 @@
   \def\textKorean{\fontspec{Noto Sans CJK KR}}
   \setmonofont{DejaVu Sans Mono}
   
+    \setsansfont{Deja Vu Sans}
+  
     \setromanfont{Times New Roman}
   
 \else
--- a/Test/expected-results/test.tex
+++ b/Test/expected-results/test.tex
@@ -17,6 +17,8 @@
   \def\textKorean{\fontspec{Noto Sans CJK KR}}
   \setmonofont{DejaVu Sans Mono}
   
+    \setromanfont{Times New Roman}
+  
 \else
   \IfFileExists{utf8x.def}%
    {\usepackage[utf8x]{inputenc}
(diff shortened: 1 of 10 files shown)


martindholmes:
In discussion with @sydb, also added Deja Vu Sans as the default Sans font.

peterstadler (commented):
I tested this locally on my MacBook and I was missing the DejaVu Sans font. I'm probably able to install this font but for really addressing issue #789 and providing sensible defaults across systems we might consider:
* to switch fonts depending on OS, or
* to check for availability of fonts on the system

martindholmes:
Good point. @sydb didn't have Times New Roman, which was a surprise to me. So I think surfacing the two font parameters and documenting them is a better approach.

martindholmes:
Good point. @sydb didn't have Times New Roman, which was a surprise to me. So I think surfacing the two font parameters and documenting them is a better approach.

martindholmes:
This doesn't seem like a practical approach.

PR #793, opened 2025-11-27
Update gitignore and Makefile

peterstadler:
This PR adds some files to the gitignore list as well as updates the Makefile `clean` target to remove/ignore files which are created during a test run by `make test`. These are all temporary files that should not be added (by accident) to version control and should be cleaned up after testing.


martindholmes (approved):
Looks good.

Issue #795, opened 2025-12-30
need to process `@key` of `<attRef>` (it is replacing `@name`)


sydb:
Stylesheets need to be able to process both `attRef/@name` and `attRef/@key` (in exactly the same way) until 2027-06-30, when support for `attRef/@name` will be dropped.

PR #796, opened 2025-12-31
Process the new `@key` of `<attRef>`

sydb:
Add processing for new `@key` attribute of `<attRef>` (in addition to, rather than replacing, now deprecated `@name`).

This is intended to address #795, itself intending to address TEIC/TEI#2850.

Passes tests locally (except PDF generation, which _always_ fails locally); P5 (using the "sydb_2850_attRef_key_not_name" branch) builds against this branch.


bleekere:
Looks good to me, passes local tests. @torstenroeder for you too? Good to merge?

PR #797, opened 2026-01-09
jTEI: split long attribute values inside eg:egXML

rvdb:
This PR inserts a zero-width space inside attribute values of elements inside `<egXML>`, before ampersands (`&`) that are _not_ followed by whitespace. 

Motivation: the FOP processor doesn't seem to consider `&` as a line-breaking character, which can cause long URLs in attribute values inside code examples to overflow the page margin. Ampersands that aren't followed by whitespace most likely occur inside URLs, so adding a zero-with space in this limited context should improve wrapping of long URLs.

See this previously merged PR: https://github.com/TEIC/Stylesheets/pull/785.

### DIFF
--- a/profiles/jtei/jtei.common.xsl
+++ b/profiles/jtei/jtei.common.xsl
@@ -244,7 +244,7 @@
           </seg>
           <seg type="abstract.egXML.attribute.value">
             <xsl:text>="</xsl:text>
-            <xsl:value-of select="local:escapeEntitiesForEgXMLAttribute(.)"/>
+            <xsl:value-of select="replace(., '&amp;([\S])', '&#8203;&amp;$1') ! local:escapeEntitiesForEgXMLAttribute(.)"/>
             <xsl:text>"</xsl:text>
           </seg>
         </xsl:for-each>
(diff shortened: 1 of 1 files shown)


profiles/jtei/jtei.common.xsl
```diff
@@ -244,7 +244,7 @@
           </seg>
           <seg type="abstract.egXML.attribute.value">
             <xsl:text>="</xsl:text>
-            <xsl:value-of select="local:escapeEntitiesForEgXMLAttribute(.)"/>
+            <xsl:value-of select="replace(., '&amp;([\S])', '&#8203;&amp;$1') ! local:escapeEntitiesForEgXMLAttribute(.)"/>
```
sydb: What do the square brackets accomplish?
That is, would `&amp;(\S)` do just as well?

sydb (approved):

profiles/jtei/jtei.common.xsl
rvdb: Ah, sure, thanks, @sydb  ! I've committed a version without the square brackets.

martindholmes (approved):

Issue #405, opened 2019-12-11
.docx to TEI P5 XML Document conversion fails


fricke-steyer:
Can you help me? Our other files are ok, only this one doesn't work. Whats wrong?
Kind regards,
Henrike

[emotion_analysis_2019.docx](https://github.com/TEIC/oxgarage/files/3863581/emotion_analysis_2019.docx)

Error occured. Please check the filetype and try again.?

Error: class pl.psnc.dl.ege.exception.ConverterException

Processing terminated by xsl:message at line 130 in fields.xsl

lb42:
Running this online with docxtotei produces the (slightly) more helpful 
error message:

  [xslt] fldSimple: unrecognized type REF BMfig_wheel \* MERGEFORMAT

which appears to relate to the reference to a graphic in section 2.2 :

"The wheel (Figure 1) is constructed in the fashion of a color wheel"

I don't have Word here, so I cannot be sure. However, if I delete that 
parenthesized reference, save the file as DOCX, and try the conversion 
again, everything works fine.

Maybe the problem is that the graphic file isn't included in the document?

TomazErjavec:
Rather than opening a new issue, I post here another Word file that causes the Stylesheets to fail. At first glance it looks easier to fix than the previous one, the error is:

```
A sequence of more than one item is not allowed as the first argument of fn:starts-with() ("VAROVALKE_1_brez ozadja copy", "VAROVALKE_2_brez ozadja copy") ; SystemID: file:/project/tei/convert/Stylesheets/docx/from/graphics.xsl; Line#: 83; Column#: 12
``` 
[TEI_Stylesheet_crash-test.docx](https://github.com/TEIC/Stylesheets/files/5370800/TEI_Stylesheet_crash-test.docx)

martindholmes:
Council F2F group looked at this and the problem is actually a pointer to something that does not exist in the Word document itself. We (me and @martinascholger and @joeytakeda)  think that in fields.xsl, at line 129, we should not terminate the processing but instead output a `<hi>` element with an error flag in the `@rend` attribute and then apply-templates to provide some helpful content.

joeytakeda:
Re this issue, @TomazErjavec: 

> Rather than opening a new issue, I post here another Word file that causes the Stylesheets to fail. At first glance it looks easier to fix than the previous one, the error is:
> 
> ```
[...]

PR #801, opened 2026-04-16
Fix for issue #405: do not fail/terminate the transformation…

martindholmes:
…, just flag the error in a message and a hi element in the output file when a w:fldSimple element cannot be successfully processed.

### DIFF
--- a/docx/from/fields.xsl
+++ b/docx/from/fields.xsl
@@ -127,7 +127,8 @@ of this software, even if advised of the possibility of such damage.
 	<xsl:when test="contains(@w:instr,'INCLUDETEXT')"/><!-- from docm to docx conversion? -->
 	<xsl:when test="contains(@w:instr,'TEMPLATE')"/><!-- from docm to docx conversion? -->	
 	<xsl:otherwise>
-	  <xsl:message terminate="yes">fldSimple: unrecognized type <xsl:value-of select="@w:instr"/></xsl:message>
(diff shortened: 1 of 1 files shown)


docx/from/fields.xsl
```diff
@@ -127,7 +127,8 @@ of this software, even if advised of the possibility of such damage.
 	<xsl:when test="contains(@w:instr,'INCLUDETEXT')"/><!-- from docm to docx conversion? -->
 	<xsl:when test="contains(@w:instr,'TEMPLATE')"/><!-- from docm to docx conversion? -->	
 	<xsl:otherwise>
-	  <xsl:message terminate="yes">fldSimple: unrecognized type <xsl:value-of select="@w:instr"/></xsl:message>
+	  <hi rend="error">fldSimple error: unrecognized type <xsl:value-of select="@w:instr"/></hi>
```
joeytakeda, line 130: In looking to see if this was the right `@rend` value to use, I learned that there is a `docxError` function (defined in docx/from/functions.xsl) that outputs a `<hi>`, a `<note>`, and an output message. We should probably use that instead of the `<hi>` and the `<xsl:message>`. So I suggest something like: 

```xml
<xsl:sequence select="tei:docxError('fldSimple error: Unrecognized type ' || @w:instr)"/>
```

joeytakeda (changes requested):

Issue #800, opened 2026-04-01
The XLink namespace should not be magically bound


sydb:
(Arising from ATOP mtg of 2026-03-31 …)

The prefix "xlink:" is handled magically in the Stylesheets. Thus an ODD writer can use 
~~~xml
<attDef ident="xlink:actuate" usage="opt">
~~~
rather than
~~~xml
<attDef ident="actuate" ns="http://www.w3.org/1999/xlink" usage="opt">
~~~~
but _only_ for this namespace using this prefix. Using a prefix does not work for anything else (except "xml:", of course). I had thought this is because the template "generateOutput" in file odd2relax.xsl hard-codes a binding of the XLink namespace to “xlink:”, but I am no longer convinced that alone is sufficient to explain this behavior.

In any case, we (the ATOP task force) believe it should not be magic. If an ODD writer wants to create (or modify) an attribute in the XLink namespace, he or she should be forced to use the `@ns` attribute as with any other namespace.

(Alternatively, we could say “an ODD writer can use any prefix they want, but it has to be declared”, but then a) there are two ways to do the same thing, b) we have to come up with a mechanism to declare a prefix (which might be just “make sure it is declared with an XML namespace declaration”, but that may be a bit fragile), and c) we have to change the Stylesheets so they parse off the prefix and look up the correct namespace for it.)

The main question we ATOPers have about this is whether or not deprecation is required, and what it would look like. So the NEEDS DISCUSSION label is mainly just for that, and if we get some thoughts here, this probably does not need to be discussed at a meeting. (I am not sure whether deprecation is appropriate or not; @HelenaSabel thinks it is not necessary, as the correct mechanism — use `ns="http://www.w3.org/1999/xlink"` — works, so whether an ODD writer gets a warning or an error it is quite an easy fix.)

P.S. I called this a “Task” not a “Bug” because it does not strike me that de-mystifying is the same as de-bugging.

bansp:
Thumbs up for demystifying.

About deprecation: the lack of it would manifest in the process of deriving schemas, I guess, with a standard error message about unbound prefix -- correct? So we're not so much talk talking about breaking an existing system for the end user, but rather breaking at the level of the potential post-release re-derivation of schemas, which should always be monitored. Seems catchable, and, as Helena points out, quite fixable, in a standard procedure.

sydb:
[...]

PR #811, opened 2026-07-08
remove magical xlink namespace binding

sydb:
Fixing #800 by removing the magic binding of the prefix "xlink:" to the namespace http://www.w3.org/1999/xlink; updated expected results to match.

The remaining question is what do we do, if anything, to warn users about this? Is this behavior documented anywhere?


sydb:
As far as I can tell nothing about "xlink" (case insensitive) is documented in any of the TEI/, Stylesheets/, or Documentation/ repos.

martindholmes:
This was reverted (perhaps temporarily) because the downstream fallout for the P5 repo was too significant to solve before the upcoming release.

Issue #809, opened 2026-05-24
HTML element thead missing in a second table


Dominique-M:
As amateur user of TEI XML (as authoring tool for HTML), I experienced that the first tei:row, if @role 'label', would induce a html:thead. Working these days on a document with two tables, I was surprised to find that the same code worked for the first table and not for the second! I went to look at html_figures.xsl and, though being still more amateur at XSL, guessed that @rend 'thead' could solve my problem, and indeed. Then, if I understand the code, what worked for the first table is the elegant solution, needing no explicit @rend 'thead', but simply not(preceding::tei:row). The problem is, if I am not mistaken, that this means no row anywhere above. To me, this does not make sense. We want to test if one table has a tei:row @role 'label' as first row of that table, therefore test that there is no preceding row in the same table, not in the whole document. Don’t we mean not(preceding-sibling::tei:row), no other row before in the same table? I let you judge if this is a bug or my not understanding XSL, Xpath, etc., and if preceding-sibling would do. I had a try (three occurrences lines 339 & ff in version 7.60.0) and to me it works, but maybe it could break something else in other situations. Same correction in the template for tei:cell, line 85. For my two tables, see under https://d-meeus.be/marxisme/classiques/Concordance.html and, for the source, the two subdivs of div xml:id="Concordance" in https://d-meeus.be/marxisme/classiques/lectures.xml.

hcayless:
This certainly seems like a bug. Your evaluation of the meaning of `preceding::tei:row` is correct. Maybe the thought process at work might have been a wish to support `<table>`'s rather goofy content model, which ought to be pruned aggressively, but it's the wrong approach. 

Given that such pruning is probably a longer-term project, I'd be in favor of having the stylesheets be more opinionated about what they'll support in a TEI -> HTML context. If someone wants to do goofy things with tables that's fine, but I don't think the stylesheets need to contort themselves to support it. I think it's ok to assume TEI tables are simple _in the context of turning them into HTML_. Really sophisticated tables are likely even now to be unable to be rendered properly in a browser. And that's ok!

sydb:
@Dominique-M (and everyone else): see #812.

PR #812, opened 2026-07-14
Enact OP’s suggested change for table headers

sydb:
Addressing #809.

1. OP’s suggested fix is exactly what jumped into my head immediately.
2. OP says that change fixes his problem.
3. With said change, Test2/ passes (at least locally on my system); Test/ failed with three diff errors, and in all three cases the actual version was better than the expected version. (So I copied actual/ to expected/ in all three cases.)


ebeshero:
This looks ready to merge, but it seems we haven't had a chance to review it before the July release.

PR #814, opened 2026-07-24
Fix spacing in ordered list template output

### LINKED ISSUE #813: TEI to Markdown: ordered lists missing whitespace
https://github.com/TEIC/Stylesheets/blob/716a76b6abef6f1556e8265e49378880a42e2f06/markdown/tei-to-markdown.xsl#L303-L311

The above template is missing one essential whitespace when creating ordered lists, right after the punctuation character just before the list item content.

```xml
<xsl:when test="tei:isOrderedList(..)">1. </xsl:when> 
```

bwbohl:
Closes #813

### DIFF
--- a/Test2/expected-results/generalP5.md
+++ b/Test2/expected-results/generalP5.md
@@ -105,13 +105,13 @@ ImageDescriptionCamera directionSVG, JPEG, GIF or PNG formatAll pictures were ta
 # Lists
 
 Various sorts of list are legal within paragraphs, and you can reference [items in lists](#birds)... 
-1.Dogs
+1. Dogs
 
-1.Zebras
+1. Zebras
 
-1.Birds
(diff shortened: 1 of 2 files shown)


sydb:
Looks fine at first glance (although I would prefer to see the space as “&amp;#x20;” in the XSLT file, just so no one misses it :​-​).
I have tested, seems to work perfectly well in both Test/ & Test2/.

HOWEVER, @bwbohl, just so you know, because there is an upcoming release in < 1 week, this should not be merged until after said release (due to our rules of engagement). Thus I have marked this as **status: pending**, pending the release next week. After that it can be merged without much further ado, as far as I am concerned.

sydb (approved):

bwbohl:
many thanks

